PR #56, opened 2017-03-09
Add tooltips to script parameters

wowselim:
In the previous version the descriptions of parameters specified in the .json-file is unused in the web ui.
This update adds tooltips to the `<input>`s (combobox, checkbox, textfield)


bugy:
Hi, thank you for the pull request. Could you please add check for null/empty description and don't set title in this case? (there is method isEmptyString in common.js)

PR #85, opened 2017-11-23
Update launcher.py to avoid the No module named 'server' error

csphoenix1:
Changing the working directory to the app's directory to avoid the "ImportError: No module named 'server'" error be begginner in Python.


bugy:
Thanks @csphoenix1, this makes perfect sense.

PR #86, opened 2017-11-27
Add some features

csphoenix1:


bugy:
Hi @csphoenix1 , thanks for the features, they look useful. I have some comments on replacing variables, please check my code review comments

src/model/model_helper.py
```diff
@@ -27,6 +28,18 @@ def is_empty(value):
...
 
+def update_output_files_with_vars(output_files, arguments):
+    for i, output_file in enumerate(output_files):
+        regex = "\$\$([a-zA-Z0-9-_]+)"  # Syntaxe is : $$my_variable (with two '$')
+        matched_arguments = re.findall(regex, output_file)
+        for argument in matched_arguments:
+            if argument in arguments:
+                value = arguments[argument]
+                output_file = output_file.replace("$$" + argument, value)
+        output_files[i] = output_file
```
bugy: Could we return modified array? Instead of changing internal state of the config?
Also I would suggest to move the function to _features/file_download_feature.py_ and call it directly inside `prepare_downloadable_files`

src/server.py
```diff
@@ -470,6 +470,8 @@ def post(self):
                 respond_error(self, 400, 'Received invalid parameters')
                 return
 
+            model_helper.update_output_files_with_vars(config.output_files, execution_info.param_values)
```
bugy: Please read my comment regarding moving the call to FinishListener and file_download_feature

src/model/model_helper.py
```diff
@@ -27,6 +28,18 @@ def is_empty(value):
     return (not value) and (value != 0) and (value != False)
 
 
+def update_output_files_with_vars(output_files, arguments):
+    for i, output_file in enumerate(output_files):
+        regex = "\$\$([a-zA-Z0-9-_]+)"  # Syntaxe is : $$my_variable (with two '$')
```
bugy: I would suggest to change internal loop a bit: we can iterate over all arguments and build regex based on it:
e.g.
```
for i, output_file in enumerate(output_files):
  for argument in arguments:
    output_file = re.sub('\$\$' + argument, arguments[argument], output_file)
```
Also secure arguments should be excluded

One more thing/question. $$ syntax is already used for environment variables in couple of places. What do you think, won't reusing the syntax cause any confusion? As for me, it shouldn't be a problem

csphoenix1: +1 about the loop.

About the syntaxe, i didn't know it was used for environment variables.
To avoid bad surprise, it would be better if we use 2 differents syntaxes. So, since the "$$" syntaxe is already used, we could use "$$$" or "$_"? What do you think about? "$$$" is not use in contrary to "$_" which is used by PHP.

src/model/model_helper.py
csphoenix1: Yes good idead, I will modified it.

src/server.py
csphoenix1: Yes no problem ;)

csphoenix1:
Also, you will have to update the Wiki :)

> You can edit conf/conf.json file, if you want to add some features to the server, of want to change default values. 

conf.json structure:
```javascript
{
  /**
    * Required: no
    * Description: custom port for running the web server
    * Type: number
    * Default: 5000 (5443 for ssl)
    */
  "port": 8080,
  /**
    * Required: no
    * Description: custom interface for running the web server
    * Type: string
    * Default: "0.0.0.0" ("127.0.0.1" or "localhost" to listen only on loopback/local or "0.0.0.0" to listen on all interfaces)
    */
  "interface": "0.0.0.0",
  /**
    * Required: no
    * Description: add ssl element to configure ssl, if needed
    * Type: object
    */
  "ssl": {
    /**
      * Required: yes
      * Description: the path to key file. Can be relative to script-server location
      * Type: string
      */
    "key_path": "testing/ssl/script-server.key",
    /**
      * Required: yes
      * Description: the path to cert file. Can be relative to script-server location
      * Type: string
      */
    "cert_path": "testing/ssl/script-server.crt"
  },
  /**
    * Required: no
    * Description: add user authentication to the server
    * Type: object
    */
  "auth": {
    /**
    * Required: yes
    * Description: type of authentication protocol. Supported types: ldap
    * ATTENTION: ldap auth requires python ldap3 module to be installed
    * Type: string
    */
    "type": "ldap",
    /**
    * Protocols: ldap
    * Required: yes
    * Description: url of the LDAP server
    * Type: string
    */
    "url": "ldap://127.0.0.1",
    /**
    * Protocols: ldap
    * Required: no
    * Default: $username
    * Description: pattern for username, which is sent to LDAP server. $username variable is filled with username from login screen
    * Type: string
    */
    "username_pattern": "uid=$username,ou=people,dc=test,dc=ldap",
    /**
    * Protocols: ldap
    * Required: no
    * Default: 3
    * Description: protocol version
    * Type: int
    */
    "version": 3
  },
  /**
    * Required: no
    * Description: alerts configuration, which is used when a script fails
    * Type: object
    */
  "alerts": {
    /**
    * Required: no
    * Description: destinations, to which an alert is sent
    * Type: array of objects
    */
    "destinations": [
      {
        /**
        * Required: yes
        * Description: type of destination. Supported types: 
        *    email: alert message is sent via email. Title and body are going to email title and body.
        *    http: alert message is sent to HTTP server via GET request. Title and body are concatenated together (separated by new line) and fill request body
        * 
        * ATTENTION: email requires smtplib python module to be installed
        * Type: string
        */
        "type": "email",
        /**
        * Destinations: email
        * Required: yes
        * Description: sender of the letter
        * Type: string
        */
        "from": "buggygm@gmail.com",
        /**
        * Destinations: email
        * Required: yes
        * Description: semicolon-separated list of recipients
        * Type: string
        */
        "to": "buggygm@gmail.com",
        /**
        * Destinations: email
        * Required: yes
        * Description: which server should be used to send an email
        * Type: string
        */
        "server": "smtp.gmail.com",
        /**
        * Destinations: email
        * Required: no
        * Default: if auth is disabled, then empty, otherwise taken from "to" field
        * Description: username to authenticate in the server
        * Type: string
        */
        "login": "buggygm@gmail.com",
        /**
        * Destinations: email
        * Required: no
        * Default: empty
        * Description: password to authenticate in the server. Environment variables can be used, for this $$ prefix should be used
        * Type: string
        */
        "password": "$$EMAIL_PWD",
        /**
        * Destinations: email
        * Required: no
        * Default: true if login or password is specified, false otherwise.
        * Description: specifies, whether we should authenticate to the server before sending an email
        * Type: boolean
        */
        "auth_enabled": true,
        /**
        * Destinations: email
        * Required: no
        * Default: true if server is gmail, false otherwise
        * Description: specifies, whether server requires TLS layer enabled
        * Type: boolean
        */
        "tls": true,
        /**
        * Destinations: http
        * Required: yes
        * Description: url to which message will be sent   
        * Type: string
        */
        "url": "localhost:5000/test_alerts"
      }
    ]
  }
}
```

src/server.py
```diff
@@ -361,6 +361,14 @@ def finished(self):
 
                 output_logger.close()
 
+                # What's the diffrence between "get_config()" and "config" ?
```
bugy: There is no difference. 
I've created a getter method because I was used to it in Java. However in python it's not needed and you can safely access class fields directly. So, please use "config.output_files"

src/server.py
```diff
@@ -361,6 +361,14 @@ def finished(self):
...
+                # process_wrapper.get_config().output_files = file_download_feature.update_output_files_vars_with_args(
+                #    process_wrapper.get_config().output_files,
+                #    process_wrapper.execution_info.param_values)#
+                process_wrapper.config.output_files = file_download_feature.update_output_files_vars_with_args(
```
bugy: could you move this call inside _prepare_downloadable_files_? So something like:
```
def prepare_downloadable_files(...):
    output_files = config.output_files

    if not output_files:
        return []
    
    output_files = update_output_files_vars_with_args(...)
```

bugy:
Regarding wiki update - thanks, I'll do it after merging your changes (both for output files and address). However I'm a bit confused - in your proposed wiki statement you used "interface" name, however according to the code it should be "address". Do I miss something? 
I'm fine with any of the names

csphoenix1:
Hi @Bugy,
I

csphoenix1:
Hi @bugy !
About interface vs address, we can finaly use "address" (according to the SSH Server Config File in Linux: they use the word "address" so we can do the same :) ).
Also, I moved the call as asked, so I hope it's that you wan't.

PS : I've created some issue about involves that it could be cool :)

bugy:
Hi @csphoenix1 , thank you for the changes and for new feature requests! :)

PR #104, opened 2018-03-22
Added startup message with url of running application

Guria:


Guria:
Hi. Yeap, could do it as well.

bugy:
Since it was my requirement, I decided to implement it myself to be fair :)
Thanks for the PR!

PR #135, opened 2018-10-02
Allow config dir and config file to be specified from command lines

andrewheberle:
This allows the currently hardcoded config dir and config files to be set on the command line.

The main driver behind this would be to have separate systemd unit files that refer to different configs.

eg

/lib/systemd/system/script-server@.service

```
[Unit]
Description=Script Server
After=network.target
StartLimitIntervalSec=0

[Service]
Type=simple
Restart=always
RestartSec=1
ExecStart=/usr/bin/python3 /opt/script-server/launcher.py --config-file %i.json

[Install]
WantedBy=multi-user.target
```

This allows different configs to be executed (maybe with different auth backends or ports or ssl config etc) as follows:

```
systemctl enable script-server@http
systemctl start script-server@http
systemctl enable script-server@https
systemctl start script-server@https

```

With matching config files:

conf/http.json

```
{
  "port": 80,
  "address": "0.0.0.0",
  "title": "Plain HTTP script server"
}
```

conf/https.json

```
{
  "port": 443,
  "address": "0.0.0.0",
  "title": "Secure HTTPS script server"
  "ssl": {
    "key_path": "conf/script-server.key",
    "cert_path": "conf/script-server.crt"
  }
}
```

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -27,7 +27,10 @@ args = vars(parser.parse_args())
 TEMP_FOLDER = 'temp'
 
 CONFIG_FOLDER = args['config_dir']
-SERVER_CONF_PATH = os.path.join(CONFIG_FOLDER, args['config_file'])
+if os.path.isabs(args['config_file']):
+    SERVER_CONF_PATH = args['config_file']
+else:
+    SERVER_CONF_PATH = os.path.join(CONFIG_FOLDER, args['config_file'])
 LOGGER = logging.getLogger('main')
 
 
--- a/src/main.py
+++ b/src/main.py
@@ -1,3 +1,4 @@
+import argparse
 import json
 import logging
 import logging.config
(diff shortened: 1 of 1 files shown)


bugy:
Hi @andrewheberle, thank you for the proposal, nice description and the implementation. 
The code looks good, the only thing I would suggest is to allow absolute config-file paths. If it's absolute, we apply it as it is, otherwise get it relative to the config folder. Could you do it please, or I can accept your PR and fix myself.

Important note regarding the general approach of running multiple servers with the same working directory: script-server was not supposed to be used like this and there would be some conflicts. For example, log files get their "unique" id based on memory counter. In this case there would be id duplicates. 

So I'll introduce a possibility to have different working directories for a script-server (based on config parameter), so all the operational files (logs, temp, etc.) will be stored there. By default configs will be read from this directory as well. What do you think?

andrewheberle:
Thanks for the feedback.

I'll update the PR to allow for absolute paths of the config file.

I did wonder about multiple instances sharing the same working directory and if there would be any conflicts, but to be honest, I was only introduced to script-server today at work, so I hadn't looked into those sort of "side effects" before firing off this PR :)

bugy:
Thank you!

PR #151, opened 2018-12-27
Fix endless while loop on EOF for pty process

jthacker:
os.read spec says that when EOF is reached by the fd, an empty bytes
object is returned.

### DIFF
--- a/src/execution/process_pty.py
+++ b/src/execution/process_pty.py
@@ -72,6 +72,8 @@ class PtyProcessWrapper(process_base.ProcessWrapper):
                     while True:
                         try:
                             chunk = os.read(self.pty_master, max_read_bytes)
+                            if not chunk:
+                                break
                             data += chunk
 
                         except BlockingIOError:


bugy:
Hi @jthacker , thanks for the PR! Looks like a critical bug. but I'm curisous why nobody saw this issue before. Let me play a little bit with a current code to check why it's working (or if I'm missing something).
After that I'll merge your PR.

jthacker:
@bugy I am running this on OSX, which might play a part there.

bugy:
The reason why it's working on linux: non-blocking read should fail with an error if there is no data to read: [example documentation](https://web.archive.org/web/20130508062559/http://www.wlug.org.nz/EAGAIN)
For OSX it should be the same behaviour (according to [this link](https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man2/fcntl.2.html), O_NONBLOCK description)

For that reason there is a code in infinite loop:
```
except BlockingIOError:
  break
```

Merging your PR should be fine, to be on safer side. But I still need to find out the reason why BlockingIOError (or any other error) is not thrown. It could be, for example, that non_blocking flag is not set at all, which will influence performance and can cause some other issues

Anyway, thank you for investigating and fixing the issue!

Issue #171, opened 2019-01-24
Docker image, automatic builds.


shinebayar-g:
It would be nice to have docker image ready.

I'll try to build docker image for the project and make it build automatically for every commit that passes test (do we have one?).

We've production ready [drone](https://github.com/drone/drone) setup that we use daily. If you don't mind I'd use that.

bugy:
Hi, thanks, nice idea. Where this image will be pushed? I don't know much about sharing docker images
Regarding the build tool - at the moment the whole process is in Travis CI: python and js tests, js build, whole package build

shinebayar-g:
I'll take a look at Travis-ci configuration if possible. Since it's an open
source project, travis ci should work.

Also since it's an open source project, our docker image should be hosted
on docker hub. I want you to make an account on docker hub, create
organization and invite me. My docker username is shinebayar. Then our
docker image will be available as under that organization's repository.

bugy:
Hi @shinebayar-g after some struggling I created an organization on docker hub and added you to a team. I believe I also have to create a repository and grant you permissions to it, right?

shinebayar-g:
Hi, I just checked docker repository. Looks like I don't have permission to create repo. Please create repository and add me. I'll test permissions to push, tag and delete.

bugy:
Added. Also I added automatic builds there: so whenever a commit to master happens, it will build :dev image based on Dockerfile. So there is no need to update any CI tool, just add a Dockerfile

shinebayar-g:
Great.

bugy:
Now images are pushed. Still to be done:
- remove dev tagging on before_deploy for non-master branches
- added versioning for stable docker images

bugy:
Added pushing docker image for stable releases as well + added incremental versioning to them
Also changed base image to slim

To be done:
- fix GitHub release, when merge with fast-forward (in this case dev release overrides stable release)
- add ":major.minor" tag (without .patch) for stable images, e.g. script-server:1.13

bugy:
### Description how to use it 
(credits go to @shinebayar-g)


Dockerfile for docker image build. Tested and working as expected.
Users should mount config files and scripts into docker image. `conf.json` file ,` runners` folder , `scripts` folder.

Example usage:

```
docker run -d -p 5000:5000 \
           -v /host/path/to/conf.json:/app/conf/conf.json \
           -v /host/path/to/runners/:/app/conf/runners \
           -v /host/path/to/scripts/:/app/scripts/ \
[...]

PR #177, opened 2019-02-02
Add dockerfile #171

shinebayar-g:
Dockerfile for docker image build. Tested and working as expected.
Users should mount config files and scripts into docker image. `conf.json`  file ,` runners` folder , `scripts` folder.

Example usage:
```
docker run -d -p 5000:5000 \
           -v /host/path/to/conf.json:/app/conf/conf.json \
           -v /host/path/to/runners/:/app/conf/runners \
           -v /host/path/to/scripts/:/app/scripts/ \
           --name script-server \
           scriptserver0/script-server:latest
```

Example runner:
```json
$ cat /host/path/to/runners/sample.json
{
  "name": "sample script",
  "description": "sample script description",
  "script_path": "/app/scripts/sample.sh"
}
```

Example script:
```shell
$ cat /host/path/to/scripts/sample.sh
#!/bin/bash
echo "Hello sample script"
```

**Notes**: Script-server users should know it cannot access host system processes as it's nature of docker container. If script-server must has access to host system processes then docker image is not suited for this kind of task. (But not saying it's impossible)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,13 @@
+FROM debian:jessie-slim
+RUN apt-get update && apt-get install -y python3 curl && rm -rf /var/lib/apt/lists/*
+RUN curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py
+RUN python3 get-pip.py && rm get-pip.py
+WORKDIR /app/
+COPY requirements.txt .
+RUN pip install -r requirements.txt
+COPY conf /app/conf/
+COPY src /app/src/
+COPY web /app/web/
+COPY launcher.py /app/launcher.py
+EXPOSE 5000
+CMD [ "python3", "launcher.py" ]


bugy:
Hi @shinebayar-g, great thanks for your efforts! Looks good for me, except one thing - as far as I see, at the moment Dockerfile just copies source files from the folder. But, for example, web directory is not source, but a build artifact.

I see 2 ways of solving this issue:
- either we pre-build files as one of the steps of Dockerfile
- or we download them from GitHub releases

In any case it will be .zip file, which contents should be extracted into docker container.

I prefer option 1, since I don't know what docker hub is capable of and don't want an extra config for it, in addition to travis.

Or, as alternative, we can try to build docker container by Travis as well, since it would have built zip file.

What do you think?

shinebayar-g:
Hi @bugy I was not sure on which files are correct ones to copy to docker image. At first I also thought I would just download latest release from github releases and extract it. But I preferred option to copy files from repository itself since dockerfile has access to it. So I downloaded 1.13.0 release and analyzed files. script-server.zip contains
```
requirements.txt
launcher.py
web
src
conf
```
files & directories. So I thought repository's matching files are correct one. But I just realized build artifacts are not included in repository, right?

Please tell me what steps are required to build correct files. Include all necessary steps to build project.

bugy:
Hi @shinebayar-g,
these 2 scripts are needed to be executed to build the script-server:
```
tools/init.py
tools/build.py
```
The first one will build JS, with the help of npm
The second one will package all required files into a build/script-server.zip file

The first step will need to download quite a lot of libraries and also requires npm installed.
The second step is needed, since it has a list of all required files, and I think it would be better to keep it in a single place (so Dockerfile will take the files from the zip file and don't care what is part of build and what is not).

pureclouds:
Hi @shinebayar-g 

My view - just load already built script server (for specific release) and use it. Other way as you suggest - load sources (for specific release) built it by npm to have web UI artifact, pack it as it done in `tools/build.py` and... at the end you will have the same as first way (load pre-built) :)

shinebayar-g:
Well both kinda works for us, let's hear opinions from @bugy . He might point some interesting case.

bugy:
Hi @shinebayar-g and @pureclouds, thanks for the suggestions!
For me preferable option would be to download already built files, to avoid any build duplication and doing the same work twice.
The only concern would be, that docker hub would need to wait until travis finishes the build and publishes the artifacts.

pureclouds:
@bugy 

>The only concern would be, that docker hub would need to wait until travis finishes the build and publishes the artifacts.

just add docker image build flow to travis (after it has built application). As [example](https://docs.travis-ci.com/user/docker/#building-a-docker-image-from-a-dockerfile) in doc.

shinebayar-g:
Indeed, since we're already using travis , there is no need of docker hub builder. Just add another step that builds docker image would be sufficient. I don't have much experience with travis-ci , I'm gonna do some experiment.

bugy:
I guess in this case we could adopt the Dockerfile either to take .zip files and extract them, or to create a separate step in travis file to extract those files to some build folder, so Dockerfile will pick them.
As for me the latter is better because of cleaner Dockerfile and no required unzip dependency. But I don't have any opinion in this regards.
@shinebayar-g what do you think?

Thanks for examples, @pureclouds

shinebayar-g:
Dockerfile can have many dependencies. Artifacts can be transferred to fresh image. Our [example](https://github.com/erxes/erxes-widgets-api/blob/master/Dockerfile).  Both ways will work, don't worry.

I should've discussed about this before push this request, consider this as done.

bugy:
Hi @shinebayar-g, your PR still makes sense, we just need to adjust a couple of lines

bugy:
Thanks for the help and advices! I've added docker pushing to travis build, please check my commit  d2a37ca67

The image name is bugy/script-server, tags are :dev for master(dev), :latest for stable
At the moment only the former one is pushed

If you would like and would have time to review my commit, I will appreciate it :)

pureclouds:
I've commented your commit with some thoughts

pureclouds:
Some thoughts about dockerization. 

I always prefer build docker images with minimal size as it possible. So, I often using alpine based base images to minimize resulting sizes. Just small example:

Dockerfile with base image `python:3.6` based on Debian:

```

**Comment by pureclouds** [comment]
Some thoughts about dockerization.

I always prefer build docker images with minimal size as it possible. So, I often using alpine based base images to minimize resulting sizes. Just small example:

Dockerfile with base image `python:3.6` based on Debian:

```
FROM python:3.6

ENV SCRIPT_SERVER_VERSION 1.13.0

WORKDIR /app

ENV PYTHONPATH="$PYTHONPATH:/app"

RUN set -e \
	&& apt-get -yq update \
	&& apt-get -yq install \
		unzip \
	&& mkdir -p /app \
	&& cd /app \
	&& wget -q https://github.com/bugy/script-server/releases/download/${SCRIPT_SERVER_VERSION}/script-server.zip \
	&& unzip script-server.zip \
	&& rm script-server.zip \
	&& pip install -r requirements.txt \
	&& apt-get -yq purge \
		unzip \
	&& rm -rf /var/lib/apt/lists/*

EXPOSE 5000

CMD ["python", "launcher.py"]
```

and Dockerfile based on `pyhton:3.6-alpine`:

```
FROM python:3.6-alpine

ENV SCRIPT_SERVER_VERSION 1.13.0

WORKDIR /app

ENV PYTHONPATH="$PYTHONPATH:/app"

RUN set -e \
	&& apk add --no-cache \
		bash \
	&& mkdir -p /app \
	&& cd /app \
	&& wget -q https://github.com/bugy/script-server/releases/download/${SCRIPT_SERVER_VERSION}/script-server.zip \
	&& unzip script-server.zip \
	&& rm script-server.zip \
	&& pip install -r requirements.txt

EXPOSE 5000

CMD ["python", "launcher.py"]
```

Now let build and see resulting docker images sizes:

```
$ docker build --no-cache -t based_on_debian -f Dockerfile . >/dev/null
$ docker build --no-cache -t based_on_alpine -f Dockerfile-alpine . >/dev/null
$ docker images --format "{{.Repository}}\t\t{{.Size}}" | grep based_on
based_on_alpine		99.7MB
based_on_debian		943MB

```

so, 10 times less. For running bash scripts Alpine well suitable I think.

**Comment by bugy** [comment]
Hi @pureclouds, thanks for the comment. I used the image from on of your comments above.
As for me I would prefer to keep debian, because if you would need to do some internal modifications/install new packages, I believe it's more popular
However we can have have a trade-off and use a slim version of the base image

**Comment by pureclouds** [comment]
Alpine based images very popular as Alpine has well maintained packages. And I think alpine more popular than debian for dockerization :)

Regarding to slim version of debian:

```
$ cat Dockerfile-slim 
FROM python:3.6-slim

ENV SCRIPT_SERVER_VERSION 1.13.0

WORKDIR /app

ENV PYTHONPATH="$PYTHONPATH:/app"

RUN set -e \
	&& apt-get -yq update \
	&& apt-get -yq install \
		unzip \
		wget \
	&& mkdir -p /app \
	&& cd /app \
	&& wget -q https://github.com/bugy/script-server/releases/download/${SCRIPT_SERVER_VERSION}/script-server.zip \
	&& unzip script-server.zip \
	&& rm script-server.zip \
	&& pip install -r requirements.txt \
	&& apt-get -yq purge \
		unzip \
	&& rm -rf /var/lib/apt/lists/*

EXPOSE 5000

CMD ["python", "launcher.py"]
```

and size:

```
$ docker build --no-cache -t based_on_debian_slim -f Dockerfile-slim . >/dev/null
$ docker images --format "{{.Repository}}\t\t{{.Size}}" based_on_debian_slim
based_on_debian_slim		165MB
```

bugy:
I decided to go with debian slim for now, since I personally use debian :) 
On docker hub it's showing 57 MB for the image size, which I think should be ok

I added mostly automated versioning. Still some improvements to be done, but I'll tackle them later in #171 

Closing the PR, since I reused the proposed Dockerfile and adopted it to the new workflow with travis
For any further discussions, please use the ticket #171 

Thanks @shinebayar-g and @pureclouds once again!

shinebayar-g:
debian-slim is fine. Also Alpine linux is great. But some binaries may not be compatible with alpin - debian changes. I'm not expert on this, so I cannot say much on this, I just use debian-slim most images are under 100mb.

PR #193, opened 2019-03-02
Add requests to requirements.txt

jthacker:
- requests module is required to support alerts

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,3 @@
+requests
 typing
 tornado<6


bugy:
Hi @jthacker , thanks for the Pull Request!
To be honest, I didn't add requests to requirements intentionally. This is an optional dependency, which is needed, only when you use HTTP alerts.
As far as I know, the best practice for python is to include only absolutely mandatory dependencies in the requirements file.

For the optional dependencies I always put a note to the doc file (though in this case it was extremely easy to overlook `requests` module there). And may be error on startup is a little bit confusing.

What do you think?

jthacker:
Good point on the optional nature of the dependency.
From a deployment standpoint it is nice to have all dependencies in one
place. What do you think about having an optional requirements file? I.e.
requirements-optional.txt or similar. At least this way, you know that if
you do pip install -r requirements.txt and pip install -r
requirements-optional.txt then all functionality in the app will work.
Alternatively using a setup.py file you can specify optional dependencies.

bugy:
Hi @jthacker, I would prefer setup.py approach. But I guess it's much more work (considering I don't have any). 
If you could add setup.py, I would really appreciate it! I'm thinking about it for a long time, but didn't have time for it

jthacker:
Unfortunately I think getting a proper setup.py is going to be a bit of work for not much payoff given that you already have a custom build system together. For now I'm going to close this request.

PR #197, opened 2019-03-19
start ScriptServer in virtual environment

cherkavi:
Just one script to download and start ScriptServer with all dependencies in current folder


bugy:
Perfect, thank you!

PR #206, opened 2019-04-09
alternative Dockerfile 

cherkavi:
with ability to create container without cloning repo. 
Package ubuntu contains all necessary packages for scripts ( scp, ssh, .... ) - with existing Dockerfile ( event with using the same network ) not possible to do any "standard" operations.

### DIFF
--- a/tools/Dockerfile_standalone
+++ b/tools/Dockerfile_standalone
@@ -1,0 +1,16 @@
+FROM ubuntu:latest
+
+
+RUN apt-get install -y python3-pip python3-dev \
+  && cd /usr/local/bin \
+  && ln -s /usr/bin/python3 python \
+  && pip3 install --upgrade pip \
+  && pip3 install wheel 
+
+
+RUN wget `curl -s https://api.github.com/repos/bugy/script-server/releases/latest | grep browser_download_url | cut -d '"' -f 4` \
+  && unzip script-server.zip \
+  && pip3 install -r requirements.txt
+
+EXPOSE 5000
+CMD [ "python3", "launcher.py" ]


bugy:
Hi @cherkavi, thanks for the PR!
Just as a little bit context: current Dockerfile is not an example/template for others. It's a part of CI process, which publishes minimal working image to a global repository. So anyone can directly run it (without building). If somebody would need extra tools, they can adjust this image for their needs (i.e. create an own Dockerfile with the base of script-server image and apt install anything extra).
For example, if you use only python scripts, any command line utilities are not needed.

On another hand, you are right, for some people these utilities are needed by default and it would make sense to have them out of the box. And having a sample Dockerfile would be helpful. 

There are a couple of things, which I think should be changed/improved:
1. I would rename the Dockerfile_standalone to Dockerfile_sample, because as for me, standalone is quite vague
2. At the same time I would rename current Dockerfile to Dockerfile_ci. So nobody will use it (this thing I'll do myself, since I have to update CI config also)
3. In your Dockerfile I see, that you set `python` command to use python3. By default on debian `python` stays for py2, and `python3` for py3. Is it different on ubuntu? I would prefer to keep this default naming, to avoid confusion for other people.

Does it make sense?

cherkavi:
Hi, @bugy - about changing name of the file - not a issue at all.
My pull request just a clue for me ( and everyone who is using ScriptServer ) to build image from scratch and don't re-invent the wheel ( I'm using it right now in OpenShift cluster ).

As you can see "my" Dockerfile can be used without cloning the repo - this is big advantage in front of existing one.

About python - Ubuntu doesn't have ( at least version 16.xx, 18.xx ) any symlink py2,py3.
Python 2 - default version ( as you know will be outdated soon ) and symlink is "python", for 3-rd version is "python3"

bugy:
Hi @cherkavi, thanks for your comments. 

> this is big advantage in front of existing one.
They are not competing, they are for different purposes. The proposed one is for other people, the existing one is only for internal usage. 

> Python 2 - default version ( as you know will be outdated soon ) and symlink is "python", for 3-rd version is "python3"
According to the proposed Dockerfile, `python` will stay for python3:
```
  && cd /usr/local/bin \
  && ln -s /usr/bin/python3 python \
```
Or do I understand /usr/local/bin and symlinks wrong?
I would suggest, that we keep default ubuntu behaviour (`python` alias/symlink for python2)

bugy:
Hi @cherkavi, may be I'm doing something wrong, but this dockerfile doesn't work for me:
1. apt-get install fails, because there is no update (and packages are not found)
2. pip3 is not working after upgrade (known issue on debian based system, with replacing system pip3 with a custom one)
3. wget/curl commands are not found

And moreover, your main point 
> Package ubuntu contains all necessary packages for scripts ( scp, ssh, .... ) 

Seems not to be true. 
```
root@7447b0f81dc0:/# less
bash: less: command not found
root@7447b0f81dc0:/# vi
bash: vi: command not found
root@7447b0f81dc0:/# vim
bash: vim: command not found
root@7447b0f81dc0:/# nano
bash: nano: command not found
root@7447b0f81dc0:/# scp
bash: scp: command not found
root@7447b0f81dc0:/# ssh
bash: ssh: command not found
```

Am I using too old or too new ubuntu image? May be some cache problem....

cherkavi:
pls, give me simple configuration ( with echo on screen and with interaction via console ) to add like an example

cherkavi:
> pls, give me simple configuration ( with echo on screen and with interaction via console ) to add like an example
( finally I've found some time to finish this work )

bugy:
Hi @cherkavi,
I couldn't do it just by configuration. So I had to create a small script
Both the script and the config attached
[guess_number.zip](https://github.com/bugy/script-server/files/3440748/guess_number.zip)

cherkavi:
> Hi @cherkavi,
> I couldn't do it just by configuration. So I had to create a small script
> Both the script and the config attached
> [guess_number.zip](https://github.com/bugy/script-server/files/3440748/guess_number.zip)

thanks a lot for the files, but seems you missed "main config file"
you just sent me one of "script" file without "main configuration"

bugy:
Hi @cherkavi, "main" configuration file is optional. I even don't know what I can put there as an example :)
Maybe that 3 configs (which are there by default):
```
{
  "port": 8080,
  "address": "0.0.0.0",
  "title": "My Script Server"
}
```

bugy:
Hi @cherkavi are you done with the changes? I mean, can I check and merge them, or are you going to add something else?

cherkavi:
I can't start server - it has empty configuration.
I have something from real life - will try to take a look on monday, or, if you have real configuration for start, can provide it for me ( your previous configuration is not enough to see something after docker start )

bugy:
Hi @cherkavi, there were some problems with paths in build/run files. I fixed them and it seems to be working.
Also, I replaced "guess_number" script with existing sample scripts. I think you can remove "guess_number" script and config from the repository

PR #268, opened 2020-03-07
Logs - replace forward slash (/) with underscore

autogun:
Hi,

I was playing with script-server for the first time today and noticed you cannot use '/' in the script name as it breaks the logging on the filesystem due to... well, obvious reasons :-)
```
FileNotFoundError: [Errno 2] No such file or directory: 'logs/processes/Group_add/remove_a_member_<host>_200307_134140.log'
```
There might be a better approach to do this as more chars may break this.

For your consideration

### DIFF
--- a/src/execution/logging.py
+++ b/src/execution/logging.py
@@ -367,7 +367,7 @@ class LogNameCreator:
         if not filename.lower().endswith('.log'):
             filename += '.log'
 
-        filename = filename.replace(" ", "_")
+        filename = filename.replace(" ", "_").replace("/", "_")
 
         return filename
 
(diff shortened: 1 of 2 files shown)


bugy:
Nice catch!
The fix looks good to me, so far after 3 years there were no complains about other chars :)

Thanks a lot!

bugy:
@autogun hmm, I had wrong tests, now they are failing because of wrongly expected name

autogun:
@bugy IMHO, back & forward slashes should be avoided when it comes to naming conventions. See if agree to the test fix :)

bugy:
@autogun ye, it's better to avoid slashes. Don't know why I used it for the test :)
Thanks for the fix!

Issue #292, opened 2020-04-15
Using auth-proxy username not detected


oktorok:
I have auth-proxy configuration, everything works fine but in the logs the variable USERNAME is None instead of the value in my user_header_name

conf.json

```
{
    "port": 5000,
    "address": "0.0.0.0",
    "title": "My Script Server",
    "max_request_size": 100,
    "access": {
	"allowed_users": "*",
	"admin_users": [ "USERA", "USERB", "USERC" ],
	"user_header_name": "X-Forwarded-User",
	"trusted_ips": "*"
    },
    "logging": {
	"execution_file": "${USERNAME}.log",
	"execution_date_format": "%y-%m-%d-%H:%M"
    }
}
```

bugy:
Hi @oktorok, thanks for reporting. It looks like a bug indeed.

oktorok:
Hi @bugy i have been looking around this bug, and although i didn't find the exact problem, because the problem has to do with request_handler and im not sure where is generated.

Anyway i can tell you that the problem appers when there are upper cases in the `user_header_name`. If in my example i put instead `x_forwarded_user` it works as expected.

bugy:
Wow, thanks for checking! I thought I knew the root cause, until you said, that it works sometimes...
Could it be, that you enabled sending basic auth credentials as well? Because in this case Script server can get username from basic auth

oktorok:
My configuration is this one:
```

{
    "port": 5000,
    "address": "0.0.0.0",
    "title": "My Script Server",
    "max_request_size": 100,
    "access": {
	"allowed_users": "*",
	"admin_users": [ "USERA", "USERB", "USERC" ],
	"user_header_name": "X-Forwarded-User",
	"trusted_ips": "*"
    },
    "logging": {
	"execution_file": "${USERNAME}.log",
	"execution_date_format": "%y-%m-%d-%H:%M"
    }
}
```

The problem i saw is that in the file `src/auth/identification.py`
```
       if new_trusted:
            if request_handler.get_cookie(self.COOKIE_KEY):
                request_handler.clear_cookie(self.COOKIE_KEY)
            if self._user_header_name:
                user_header = request_handler.request.headers.get(self._user_header_name, None
)
```
`self._user_header_name` instead of be "X_Forwarded_User" is i specified in conf file, was "x_forwarded_user" which obviously was not in request headers... 

[...]

PR #293, opened 2020-04-17
Bug user_header_name

oktorok:
Modified `identification.py` and `audit_utils.py` as said in #292.

### DIFF
--- a/src/auth/identification.py
+++ b/src/auth/identification.py
@@ -15,6 +15,10 @@ class Identification(metaclass=abc.ABCMeta):
     def identify(self, request_handler):
         pass
 
+    @abc.abstractmethod
+    def identify_for_audit(self, request_handler):
+        pass
+
 
 class AuthBasedIdentification(Identification):
     def __init__(self, authentication_provider) -> None:
@@ -24,8 +28,10 @@ class AuthBasedIdentification(Identification):
         current_user = self._authentication_provider.get_username(request_handler)
         if not current_user:
             raise Exception('Not authenticated')
-
         return current_user
+
+    def identify_for_audit(self, request_handler):
+        return self.identify(request_handler)
 
 
 class IpBasedIdentification(Identification):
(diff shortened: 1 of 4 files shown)


bugy:
Oops, there are some tests failing now, could you have a look please?
https://travis-ci.org/github/bugy/script-server/builds/676418652

oktorok:
I don't understand why they are failing, all the 23 errors are the same:
```
AttributeError: '' object has no attribute 'identification'
```
When calling `request_handler.application.identification.identify_for_audit(request_handler)` however if i execute the same tests they don't raise any error.
:(

bugy:
Hi, I think the problem is in `ip_idenfication_test.py` line 15: request_handler mock doesn't specify `identification`, only `auth`

It might be the same for other tests

Regarding running tests locally: I couldn't reproduce it either. Most probably IP resolving works differently on travis and the tests goes into a bit different flow, where identification is required

oktorok:
In `identification.py`, in `AuthBasedIdentification`
```
    def identify(self, request_handler):
        current_user = self._authentication_provider.get_username(request_handler)
        if not current_user:
            raise Exception('Not authenticated')
```
When `current_user` is `None` it raise an error, blocking some tests than before it returned none to `get_all_audit_names`, what whould be better? change the tests? or change identify? 

I think raise that error is important maybe change `identify_for_audit` for not call `identify`:
```
def identify_for_audit(self, request_handler):
        current_user = self._authentication_provider.get_username(request_handler)
        return current_user
```
What do you think?

bugy:
Hi @oktorok, I see you already made some tries, thanks a lot for your effort and sorry for that!

Regarding the fix, I think changing the test would be better. I would prefer keeping production code behaviour and fail if something goes really wrong.

oktorok:
Haha don't worry, i'm using this for learn OOP and actually is always have found debugging pretty fun.

Regarding the test, then now the correct behavior when identifying is raise Authentication error in some cases.
Then for change the tests, they should accept the error as the correct behavior? how do you specify to travis that an error is what he should expect?

bugy:
There is a line above your changes
```
handler_mock.application.auth.get_username = lambda x: auth_username
```
As you can see, I'm not using real authenticator here, but just mock a simple one
The same you could do for your identification, instead of using real `AuthBasedIdentification`, you can define your custom identify method, e.g.
```
handler_mock.application.identification = mock_object()
handler_mock.application.identification.identify_for_audit = lambda x: auth_username
```
And actually I think you can safely remove everything related to `handler_mock.application.auth` in the `mock_request_handler` method

oktorok:
Yeeeeeeyyy!!!

PR #297, opened 2020-04-24
chore(docs): Fix admin interface screenshots link

playpauseandstop:
Fix link in README, which previously lead to 404 page, instead of actual wiki page with screenshots.

----

Looks like GitHub understands relative links as paths inside the repo, which results in attempt to open `https://github.com/bugy/script-server/blob/master/bugy/script-server/wiki/Admin-interface` and not the actual wiki page with screenshots

Screen Shot 2020-04-24 at 20 30 28

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ No script modifications are needed - you configure each script in Script server 
 
 [DEMO server](https://script-server.net/)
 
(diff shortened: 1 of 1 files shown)


bugy:
Oh, thanks a lot!

PR #305, opened 2020-06-04
Gitlab oauth support

miksir:
Support of gitlab authentication
https://github.com/miksir/script-server/wiki/Authentication#gitlab-oauth


bugy:
Wow, that's mighty! thanks a lot! 

Just a couple of questions before I start digging into the code:
#### 1
> dump - filename where to persist user/group information
> session_expire_min - user inactivity time before logged out

do you think it's GitLab specific? Or it can be re-used for all authentications?

#### 2
> dump

does it have default value? I think it would be ok to save it to `temp` folder by default

#### 3
Do you store user information anywhere, to perform additional requests (for `ttl`, for example). I tried avoiding to store any user information, to reduce user credentials (or tokens) exposure. 
So if someone gets access to Script server or the user cookie, they still won't be able to access 3rd party auth services (i.e. google, gitlab or ldap)

miksir:
> do you think it's GitLab specific? Or it can be re-used for all authentications?
both implemented only for gitlab auth at this moment
`session_expire_min` can be useful for all auth providers
`dump` - depend on where we want to to store some session data... I decided to store in memory and dump to file, but may be someone will decide to use secure cookie or database :)
if your question about naming - its ok to rename

> does it have default value? I think it would be ok to save it to temp folder by default
nope, if `dump` missed - it's totally disabled and restart of app will logout all users

> Do you store user information anywhere, to perform additional requests (for ttl, for example). I tried avoiding to store any user information, to reduce user credentials (or tokens) exposure.
So if someone gets access to Script server or the user cookie, they still won't be able to access 3rd party auth services (i.e. google, gitlab or ldap)
Well, for sure it's stored in memory
`dump` store it in file, but this options can be skiped

Also I think to implement special type of authentication without group support, in this case I can ask read_user scope from gitlab, it's read only, so user api token can be used only for reading some information. 

I will think how to hide data in dump file. May be to encrypt it with key passed via ENV. But this problem is wider... for example, how to stop accessing logs because user can remove logs with script.

miksir:
Added information about `dump` to wiki page

miksir:
> What kind of user data/credentials do you store in the moment?

Only OAuth2 token but with API scope, so it's a lot. There is `read_user` scope, its safe, but can't fetch user's groups with it.

> By the way, did you consider storing it in user secured cookies?

well, it's possible, can do it

> I think if the server is compromised, there is not much you can do about server data itself. 

I am thinking about "script wrapper" for run script with sudo or in docker container. It can help to restrict access to script server files.

miksir:
Gitlab oauth key now stored in secure cookie. For better safety now it's possible to auth with `read_user` gitlab scope - it's read only scope with very limited access, but gitlab groups will not be used.

src/auth/auth_gitlab.py
```diff
@@ -0,0 +1,339 @@
...
+        if self.session_expire:
+            for userData in list(self.user_states.keys()):
+                if (self.user_states[userData]['visit'] + self.session_expire) < now:
+                    LOGGER.debug("User %s session expired and removed" % userData)
+                    del self.user_states[userData]
+
+    def dump_sessions_to_file(self):
+        if self.gitlab_dump:
+            dumpFile = open(self.gitlab_dump, "w")
+            dumpFile.write(escape.json_encode(self.user_states))
```
bugy, line 217: I think it's better use `with open...` syntax here, otherwise you can end up with open file descriptor, if any exception happens during serialisation
You can also use `file_utils.write_file(escape.json_encode(self.user_states))` here

src/auth/auth_gitlab.py
```diff
@@ -0,0 +1,339 @@
...
+        if self.user_states.get(user) is None:
+            return []
+        if self.user_states[user]['groups'] is None:
+            return []
+        return self.user_states[user]['groups']
+
+    def logout(self, user, request_handler):
+        request_handler.clear_cookie('token')
+
+    def clean_expired_sessions(self):
```
bugy, line 206: Maybe you can call this guy from some timer?

src/auth/auth_gitlab.py
```diff
@@ -0,0 +1,339 @@
...
+    def dump_sessions_to_file(self):
+        if self.gitlab_dump:
+            dumpFile = open(self.gitlab_dump, "w")
+            dumpFile.write(escape.json_encode(self.user_states))
+            dumpFile.close()
+            LOGGER.debug("Dumped state to file %s" % self.gitlab_dump)
+
+    def do_update_user(self, user, access_token):
+        self.user_states[user]['updating'] = True
+        tornado.ioloop.IOLoop.current().spawn_callback(self.update_user_state, user, access_token)
```
bugy, line 223: Out of curiosity: why are you doing it in a separate callback?

src/auth/auth_gitlab.py
```diff
@@ -0,0 +1,339 @@
...
+        else:
+            LOGGER.info("Groups for %s refreshed: " % user + str(group_list))
+            self.user_states[user]['groups'] = group_list
+        now = time.time()
+        self.user_states[user]['updating'] = False
+        self.user_states[user]['updated'] = now
+        self.user_states[user]['visit'] = now
+        self.clean_expired_sessions()
+        self.dump_sessions_to_file()
+        return
```
bugy, line 244: I think the return is redundant

src/auth/auth_gitlab.py
```diff
@@ -0,0 +1,339 @@
...
+
+        groups = []
+        for group in group_list:
+            if group.get('full_path'):
+                groups.append(group['full_path'])
+
+        return groups
+
+    @gen.coroutine
+    def validate_user(self, code, request_handler):
```
bugy, line 291: I would suggest to call it fetch/load user (or anything similar). As far as I can understand from the code. the main purpose of the function is loading the data, not data validation

src/auth/auth_gitlab.py
```diff
@@ -0,0 +1,339 @@
...
+                LOGGER.error(error_message)
+                raise AuthFailureError(error_message)
+
+        LOGGER.info("User %s group list: " % user_response['email'] + str(user_groups))
+        user_response['groups'] = user_groups
+        user_response['updated'] = time.time()
+        user_response['visit'] = time.time()
+        user_response['updating'] = False
+        oauth_access_token = user_response.pop('access_token')
+        oauth_refresh_token = user_response.pop('refresh_token') # not used atm
```
bugy, line 319: If it's not used, could you remove it, please? Or you are going to add it later in PR?

src/auth/auth_gitlab.py
```diff
@@ -0,0 +1,339 @@
...
+            if user_groups is None:
+                error_message = 'Cant read user groups'
+                LOGGER.error(error_message)
+                raise AuthFailureError(error_message)
+
+        LOGGER.info("User %s group list: " % user_response['email'] + str(user_groups))
+        user_response['groups'] = user_groups
+        user_response['updated'] = time.time()
+        user_response['visit'] = time.time()
+        user_response['updating'] = False
```
bugy, line 317: Don't you miss "state" field?

src/auth/auth_gitlab.py
```diff
@@ -0,0 +1,339 @@
...
+        now = time.time()
+        self.user_states[user]['updating'] = False
+        self.user_states[user]['updated'] = now
+        self.user_states[user]['visit'] = now
+        self.clean_expired_sessions()
+        self.dump_sessions_to_file()
+        return
+
+    @gen.coroutine
+    def update_user_state(self, user, access_token):
```
bugy, line 247: What does it method supposed to do? What kind of data does it load?

src/auth/auth_gitlab.py
```diff
@@ -0,0 +1,339 @@
...
+
+    def logout(self, user, request_handler):
+        request_handler.clear_cookie('token')
+
+    def clean_expired_sessions(self):
+        now = time.time()
+        if self.session_expire:
+            for userData in list(self.user_states.keys()):
+                if (self.user_states[userData]['visit'] + self.session_expire) < now:
+                    LOGGER.debug("User %s session expired and removed" % userData)
```
bugy, line 211: I think you can change it to info. It won't happen so often, but could help with investigating issues.

src/auth/auth_gitlab.py
```diff
@@ -0,0 +1,339 @@
...
+            return False
+        access_token = access_token.decode("utf-8")
+
+        if self.user_states.get(user) is None:
+            LOGGER.debug("User %s not found in state" % user)
+            return False
+
+        if self.user_states[user]['state'] is None or self.user_states[user]['state'] != "active":
+            LOGGER.info("User %s state inactive: " % user + str(self.user_states[user]))
+            del self.user_states[user]
```
bugy, line 169: I would suggest, that reading method (as for me, `is_smth` is a reading method) shouldn't have any side effects and modify any data. What do you think?
Also, if the user is not active, we are doing logout anyway

Maybe we could just use another name. Actually I think "validate" user might be better fit here :)

src/auth/auth_gitlab.py
```diff
@@ -0,0 +1,339 @@
...
+
+        gitlabPrefix = params_dict.get('url')
+        if not model_helper.is_empty(gitlabPrefix):
+            self._GITLAB_PREFIX = gitlabPrefix
+
+        self.states = {}
+        self.user_states = {}
+        self.gitlab_update = params_dict.get('auth_info_ttl')
+        self.gitlab_dump = params_dict.get('state_dump_file')
+        self.gitlab_group_support = params_dict.get('group_support', True)
```
bugy, line 126: Do you think there are use cases when we don't want to load GitLab groups?

src/auth/auth_gitlab.py
```diff
@@ -0,0 +1,339 @@
...
+        return self.validate_user(code, request_handler)
+
+    def is_active(self, user, request_handler):
+        access_token = request_handler.get_secure_cookie('token')
+        if access_token is None:
+            return False
+        access_token = access_token.decode("utf-8")
+
+        if self.user_states.get(user) is None:
+            LOGGER.debug("User %s not found in state" % user)
```
bugy, line 164: Isn't it a warning? Sounds like exceptional behaviour

src/tests/auth/test_auth_gitlab.py
```diff
@@ -0,0 +1,202 @@
...
+                "visit": now-10,
+                "id": 2,
+                "username": "nogroups",
+                "name": "John",
+                "email": "nogroups@test.com",
+                "state": "blocked"
+            }
+        }
+
+        os.write(tmp[0], str.encode(escape.json_encode(state)))
```
bugy, line 60: Maybe you could use test_utils here? There is `test_utils.temp_folder` and `test_utils.create_file`

Also in the test class you do 
```
    def setUp(self):
        test_utils.setup()

    def tearDown(self):
        test_utils.cleanup()
```
And in this case the data will always be cleaned up after execution. At the moment, if any assertion fails, you won't close the temp file.

src/tests/auth/test_auth_gitlab.py
```diff
@@ -0,0 +1,202 @@
...
+        os.close(tmp[0])
+        os.unlink(tmp[1])
+
+    @patch('time.time', mock_time)
+    @patch('auth.auth_gitlab.GitlabOAuthAuthenticator.do_update_user', mock_do_update_user)
+    @patch('auth.auth_gitlab.GitlabOAuthAuthenticator.do_update_groups', mock_do_update_groups)
+    def test_gitlab_oauth_user_read_scope(self):
+        now = time.time()
+
+        state = {
```
bugy, line 152: I think the previous test method has pretty similar state/config. May be you could extract common part to a separate variable/method?

web-src/src/login/login.js
```diff
@@ -84,6 +86,36 @@ function setupGoogleOAuth(loginContainer, authConfig) {
     processCurrentOauthState();
 }
 
+function setupGitlabOAuth(loginContainer, authConfig) {
+    var credentialsTemplate = createTemplateElement('login-gitlab-template');
+    loginContainer.appendChild(credentialsTemplate);
+
+    var oauthLoginButton = document.getElementById('login-gitlab-button');
+    oauthLoginButton.onclick = function () {
```
bugy, line 94: I think this onclick handler is exactly the same as for google oauth. Could you extract it to a common function, please?

src/web/server.py
```diff
@@ -171,7 +174,10 @@ def wrapper(self, *args, **kwargs):
 
 
 def has_admin_rights(request_handler):
-    user_id = _identify_user(request_handler)
+    try:
+        user_id = _identify_user(request_handler)
+    except Exception:
```
bugy, line 179: Could you logging for the error, please?

src/tests/auth/test_auth_gitlab.py
```diff
@@ -0,0 +1,202 @@
...
+mock_do_update_groups = Mock()
+mock_do_update_user = Mock()
+mock_request_handler = Mock(**{'get_secure_cookie.return_value': "12345".encode()})
+
+
+class TestAuthConfig(TestCase):
+    @patch('time.time', mock_time)
+    @patch('auth.auth_gitlab.GitlabOAuthAuthenticator.dump_sessions_to_file', mock_dump_sessions_to_file)
+    @patch('auth.auth_gitlab.GitlabOAuthAuthenticator.do_update_groups', mock_do_update_groups)
+    def test_gitlab_oauth(self):
```
bugy, line 32: To be honest I think this test method does too many things. Can we split it into multiple tests with separate assertions? 
I think as a rule of thumb, all test methods should call tested code only once and then perform assertions. For every call, there should a new test_method

What do you think?

bugy:
Hi @miksir, I checked the PR. At the moment there is a lot of duplication between Google oauth and GitLab oauth. What do you think about extracting common part to some generic OauthAuthenticator? And then let google and GitLab implementations implement only the part, which is different?

I think it would be fine if you move your new logic (with logout, dump, ttl, etc.) to this common class as well, so it will be used by google authenticator also.

src/auth/auth_gitlab.py
miksir: May be, but it's really hard for me to find how to set timer in tornado event loop. I am new in python and tornado :) And there is no big advantages in timer. Just free some memory but size of `user_states` never be more than total number of users. And you should be prepared for the situation that all users online at a time.

src/auth/auth_gitlab.py
miksir: ok

src/auth/auth_gitlab.py
miksir: Pros: User action not blocked while updating user info because it can take time, especially for cloud Gitlab
Cons: If gitlab access revoked user can do some number of requests before logged out
But I decided that pros is more important - lags of each auth_info_ttl seconds is not good

src/auth/auth_gitlab.py
miksir: ok

src/auth/auth_gitlab.py
miksir: agree

src/auth/auth_gitlab.py
miksir: Yep, I can remove. I just dont known if refresh token really works for gitlab. Anyway any refreshing of expired token not implemented, so can remove atm

src/auth/auth_gitlab.py
miksir: `state` coming from gitlab answer https://docs.gitlab.com/ee/api/users.html#for-user

src/auth/auth_gitlab.py
miksir: This is request of user information. Same as at login point. Called `update_user_state` because we are only interesting in `state` field

src/auth/auth_gitlab.py
miksir: ok

src/auth/auth_gitlab.py
miksir: Yep, I will rename it to validate or validate_user

src/auth/auth_gitlab.py
miksir: As I described in doc, there is big difference in oauth scopes. Scope - it's type of access to different endpoints in API.

Safest scope is "read_user" - it's readonly access to only one endpoint -user info. But this scope is not enough to get group list. 

Second scope is `api` - its read/write access to all API. It's allow to get groups but in case of token leak - gives much more access to Gitlab.

So I decided to give end user the choice - groups or safe token.

Oh, just found that new scope `read_api` introduced in fresh version. Probably need new option :)

src/auth/auth_gitlab.py
miksir: There is few leagal cases. For example - restart app without state_dump_file option. Or some concurrency when we are logged out user (and asked browser to remove cookie) but few HTTP request  already in progress with valid cookie.

src/tests/auth/test_auth_gitlab.py
miksir: Ok

src/tests/auth/test_auth_gitlab.py
miksir: Not exactly the same, there is difference. In case of 3-4 methods may be its ok to move base value to separate method and extend it in test, but in this case its just extra complexity

web-src/src/login/login.js
miksir: Ok

src/web/server.py
miksir: Ok, but I found that it is legal case. /auth/info is not require authenticated user, but in case of requesting without authentication its gives 500.
May be it's just a not good place to fix and better to move catch to AuthInfoHandler?

src/tests/auth/test_auth_gitlab.py
miksir: I agree but just selected the shortest way :) I will try to rewrite tests but not sure how much time it can takes

miksir:
Both is OAuth, but lot of small difference. logout, dump, ttl needs only because user validity recheck implemented, but in Google OAuth not. I see what you want but it's require massive refactoring of Authenticator :(

bugy:
>  but lot of small difference. logout, dump, ttl needs only because user validity recheck implemented

But all of those can be used for google as well, cannot they? The only difference is how you parse the response. For Gitlab you check the state (and probably groups), but for google you just get email field. 
If there is an error or email is missing, the use is also inactive and can/should be logged out

bugy:
Hi @miksir should I try to help you with the refactoring? I should have time on Thursday, I think.

miksir:
I pushed some changes. Please, check unresolved with my answers - do you need more info? I will try to refactor unit tests today.

If you want to refactor google oauth too - yes, I need you help because I has other priority tasks from users.

bugy:
Hi @miksir, I'm working on the refactoring, hope to finish until end of the day
Are you going to do any changes here, today? Because it would be quite hard to merge them

miksir:
Nop, no changes planned.  But I rebased branch (no changes, just rebase), so rebease too :)

bugy:
Hi @miksir I did common behaviour extraction to a separate class. Unfortunately, I couldn't test Gitlab auth, so if you could check (and probably fix it), it would be cool.

I'll still work on unit tests

I also changed current behaviour a bit:
1. I'm not storing user state anymore. If it becomes disabled, I just remove user's state completely and he has to re-login. I thought, that storing a state was a bit redundant
2. A user is validated only during his `validate_user` phase and only this user
3. Access_token is stored only if auth_info_ttl is enabled
4. Dump file is taken every 30 seconds, to avoid potential clashes when multiple users are being updated simultaneously. I think, I'll also add checksum checking, to avoid unnecessary writes, if nothing has changed
5. Custom OAuth implementations have to define `fetch_user_info` and `fetch_groups` (i.e. how to load the data). Everything else is managed by a common class
6. I changed user state from a dictionary to a dedicated class. For me, it's easier to work in PyCharm, when the IDE knows about existing fields and can suggest/validate them :)
7. The dump file structure was changed a bit, to address 1. and 6., so I believe you would have to delete the dump file before testing

I think I went to deep and probably you won't recognize your code anymore :sweat_smile: But what do you think about the new flow?

miksir:
> 1. I'm not storing user state anymore. If it becomes disabled, I just remove user's state completely and he has to re-login. I thought, that storing a state was a bit redundant

> 4. Dump file is taken every 30 seconds, to avoid potential clashes when multiple users are being updated simultaneously. I think, I'll also add checksum checking, to avoid unnecessary writes, if nothing has changed

Sorry, I am confused :) Dump and storing user state was the same :) It was dump of users state.

bugy:
I meant state['state'] :) Like 'active', 'error', etc.

miksir:
I C. Great work. It's look like you can refactor ldap auth too :)

bugy:
Ye, it would be nice... But i think it's too much for now :)
One question: for session expiry you chose minutes, do you have a real case
for it? Usually it's days. I would think that hours would be more than
enough
And in this case, it would be way easier to implement it, by setting expiry
time for the username cookie

miksir <notifications@github.com> schrieb am Fr., 12. Juni 2020, 23:03:

miksir:
> One question: for session expiry you chose minutes, do you have a real case for it? Usually it's days. I would think that hours would be more than enough

Well, may be someone want 30 min session.  So just got minimal usable. But, imho, best way is to use human readable form. Like 30m or 1h30m

bugy:
Hi @miksir I'm done with the refactoring.

Let's keep the session behaviour as per your implementation for now: I realized, that using cookie expire time would have different behaviour: the time wouldn't be reset on user actions and would just measure time after login.

If gitlab auth is still working for you, and you are fine with my changes, we can merge the PR

miksir:
Will check today. But i think you can register on gitlab.com and add application here https://gitlab.com/profile/applications for test

bugy:
Alright, I finally tested it and it seems to work :)
Thanks a lot @miksir for the great work!

PR #306, opened 2020-06-04
secret_storage_file config param for store cookie secret

miksir:
`secret_storage_file` option for move secure cookie key file. Otherwise cookie reseted on docker container recreation.

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -40,8 +40,7 @@ else:
 LOGGER = logging.getLogger('main')
 
 
-def get_secret(temp_folder):
-    secret_file = os.path.join(temp_folder, 'secret.dat')
+def get_secret(secret_file):
     if os.path.exists(secret_file):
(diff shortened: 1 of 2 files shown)


bugy:
Hi @miksir thanks for your changes, they definitely make sense!
But what do you think about configuring the whole temp folder instead? I think it would be more flexible/helpful

miksir:
IMHO no. TEMP folder is for temporary files which is safe to delete. Temporary files should not change application behavior if removed. But files with any kind of state - it's not temporary file. In point of linux directories view - it's or /etc (config) or /var. IMHO it's much closer to config, many frameworks even has config option "secret key" for hmac signing.

bugy:
Hmm, maybe Script server should store the secret file in the config folder then?

miksir:
May be :) as for me - separate path is better because conf folder mounted as ro volume to container

bugy:
ok :)

PR #307, opened 2020-06-04
Empty allowed users

miksir:
Well... not sure about this fix, but if someone uncheck 'Allow all' but do not enter any username - empty list created and script hided from all users.

### DIFF
--- a/src/model/script_config.py
+++ b/src/model/script_config.py
@@ -239,7 +239,7 @@ def read_short(file_path, json_object):
     config = ShortConfig()
 
     config.name = _read_name(file_path, json_object)
-    config.allowed_users = json_object.get('allowed_users')
+    config. allowed_users= json_object.get('allowed_users')
     config.group = read_str_from_config(json_object, 'group', blank_to_none=True)
 
     hidden = read_bool_from_config('hidden', json_object, default=False)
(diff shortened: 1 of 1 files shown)


bugy:
Hi @miksir, to be honest, I left this on purpose. Maybe, you are right, and it doesn't make sense: most probably people would expect different behavior. But I would suggest fixing it on UI (i.e. if no users are not specified, set the property as null, rather than empty list). And keep the possibility to use empty list in a json file.

miksir:
Agree. I thought about fix UI but it's little bit more complicated for me.
But can't see the point why admin want to hide script from self.
So, may be admins should see all scripts, or may be always to add creator to allowed_users.

bugy:
> Agree. I thought about fix UI but it's little bit more complicated for me.

If you have frontend utilities installed, you can edit ScriptConfigForm.vue:152

> But can't see the point why admin want to hide script from self.
> So, may be admins should see all scripts, or may be always to add creator to allowed_users.

This might be true. There are not many reasons to have an empty list in the configuration. But as for me, being explicit here is ok. If you want to prohibit full access (for whatever reason): use empty list (in the json file), if you want to leave access for admins, use `"allowed_users": ["@admin_users"]`

miksir:
Well, really need to fix frontend. This tags thing is confuse users a lot. Will reopen this MR soon.

Issue #288, opened 2020-04-03
Option for put no space between parameter and value


IncredibleRichie:
To execute packer with the parameter
`-var-file=`

you need to put the path directly behind 
'='
with no space.

When i select a server-file it puts a whithespace between the parameter and the path.

At the moment i think this is not posible to change the behavior.
If it is can u tell me how? 

btw. Thanks for your great work.

bugy:
Hey, unfortunately, it's not supported :(
For some reason, nobody ever asked about it (at least I couldn't find a ticket)

IncredibleRichie:
I looked for it too becauce i could not  belive it. ^^

bugy:
From #308 
@miksir wrote:
> Hi! I got a request for support --arg=value. Currently there is no options to do it because space added between argument name and value. I can see two ways - separate options for it (do_not_add_space \o/) or transparent behaviour - if arg name ended with = - do not include space. What you think about second way?

Hi @miksir I think the second one is cooler, though a bit dangerous. I think we can try it and see if there are any complaints :)

miksir:
ok, will make PR today

yosefy:
example:
 terraform apply -var="image_id=ami-abc123"

bugy:
Hi @yosefy, how do you do it now? With a wrapping script?

miksir:
Is quotes required here? Usually quotes used only if value with spaces and only because shell use space as arg delimiter. But in case of popen args passed to system as separate items and quotes not required. Am I right?

PR #310, opened 2020-06-10
#288 Remove space between arg name and value, support of arg repeating with each value

miksir:
Issue #288


bugy:
Or may be be same_arg_param

miksir:
Renamed

bugy:
There are 2 web tests failing, would you mind having a look? Or should I?

miksir:
Oh, I will check.

miksir:
Probably fired because I changed visibility of some fields when they are unused (they was disabled before)

PR #316, opened 2020-06-18
Two issue has been added:

jtprof:
1. --log-folder and --tmp-folder command line arguments to get the files locations more flexible
2. "expiration_days" (non-intеger values are avalible also) in server config file's section "auth" to limit auth cookie life time. The default value is 30 days if the param is underfined.

 Changes to be committed:
	modified:   src/auth/auth_base.py
	modified:   src/auth/tornado_auth.py
	modified:   src/main.py
	modified:   src/model/server_conf.py

### DIFF
--- a/src/model/server_conf.py
+++ b/src/model/server_conf.py
@@ -156,6 +156,8 @@ def create_authenticator(auth_object, temp_folder):
     else:
         raise Exception(auth_type + ' auth is not supported')
 
+    authenticator.auth_expiration_days = float(auth_object.get('expiration_days')) if auth_object.get('expiration_days') is not None else 30
+
     authenticator.auth_type = auth_type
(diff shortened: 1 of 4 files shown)


src/model/server_conf.py
```diff
@@ -156,6 +156,8 @@ def create_authenticator(auth_object, temp_folder):
     else:
         raise Exception(auth_type + ' auth is not supported')
 
+    authenticator.auth_expiration_days = float(auth_object.get('expiration_days')) if auth_object.get('expiration_days') is not None else 30
```
bugy, line 159: I think you can use default value here:
`auth_object.get('expiration_days', '30')`

bugy:
Nice, thanks!

Issue #217, opened 2019-05-20
allow auth.username and auth.audit_name in script values


muzzol:
opening this feature request as recommended here: #130 

it would be nice to allow auth variables in script values.
ex:

```
    {
      "name": "Roles",
      "type": "list",
      "values": { "script": "/opt/scripts/list_rols.sh ${auth.audit_name}" }
    },


```

autogun:
Hey @bugy,

I'd like to add this one, also to have ${auth.username} templated in the `description` field.
Can you point me to the right place?

bugy:
Hi @autogun, for filling the values, there is a `model_helper.replace_auth_vars` method

For script substitution: parameter_config.py:183
For the description: parameter_config.py:70

PR #317, opened 2020-06-22
allow auth.username and auth.audit_name in script values and descriptions

autogun:
This an attempt to support https://github.com/bugy/script-server/issues/217.
I also thought having those variables templated in script's config `description` field aswell.

### DIFF
--- a/src/model/model_helper.py
+++ b/src/model/model_helper.py
@@ -189,7 +189,9 @@ def fill_parameter_values(parameter_configs, template, values):
 
 def replace_auth_vars(text, username, audit_name):
     result = text
-
+    if not result:
+        return
+        
     if not username:
         username = ''
     if not audit_name:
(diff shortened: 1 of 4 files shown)


bugy:
Hi @autogun thanks for your changes!
Why did you decide to use "resolve_default" instead of "replace_auth_vars"? To be honest, I think now the naming is a bit confusing

autogun:
Hey @bugy,

You mean as simple as `self.description = replace_auth_vars(config.get('description'))`? 😄

bugy:
Exactly (you would also need to path auth arguments). And the same for the script value

src/model/model_helper.py
```diff
@@ -189,7 +189,9 @@ def fill_parameter_values(parameter_configs, template, values):
 
 def replace_auth_vars(text, username, audit_name):
     result = text
-
+    if not result:
+        return
```
bugy, line 193: Please `return result`

bugy:
Thanks!

PR #319, opened 2020-06-30
dependant list added. 

jtprof:
The values in the list, which are available to a user choice, depends on chosen values of other parameters. 
The relations should be set in configuration manually. 
Parameter configuration window in admin has not been changed.

The feature useful to avoid user to set inadmissible values combination.  

Example in parameterized.json:
'aduld dogs food' in "Dependant List" variable is unavailable while "Cats", "Rats" or "Squirrel" is chosen in "Simple List" variable.


jtprof:
A far as i have seen CI test failed due to f-string syntax (PEP 498). 
May be it's good idea to increase project's requirements to Python 3.6 ?

src/config/constants.py
```diff
@@ -1,4 +1,6 @@
 PARAM_TYPE_SERVER_FILE = 'server_file'
 PARAM_TYPE_MULTISELECT = 'multiselect'
+PARAM_TYPE_LIST = 'list'
+PARAM_TYPE_DEPENDANT_LIST = 'dependant_list'
```
bugy, line 4: I think this is not exactly "param type" because the type is still "list" (according to the example, you added)

samples/configs/parameterized.json
```diff
@@ -31,11 +31,34 @@
...
+        "Dogs",
+        "Rats",
+        "Squirrel"
       ]
     },
+    {
+      "name": "Dependant List",
+      "param": "--dependant_list",
+      "type": "list",
+      "description": "The avalible values depends on other variables",
```
bugy, line 44: Could you fix typos please? :) "Available" "depend"

src/config/script/list_values.py
```diff
@@ -52,6 +52,46 @@ def __init__(self, script) -> None:
...
+
+    def get_values(self, parameter_values):
+        self._availible_values = set()
+        deny_values = set()
+
+        for e in self._values_conditions:
+            for key in self._values_conditions[e]:
+                pvalue = parameter_values.get(key)                
+                if not is_empty(pvalue):
+                    if not(pvalue in self._values_conditions[e][key] ):
```
bugy, line 86: could you change it to `not in`, please?

src/config/script/list_values.py
```diff
@@ -52,6 +52,46 @@ def __init__(self, script) -> None:
...
+            raise Exception(f"Exception {ex} during process{self._var_name}")
+
+    def get_required_parameters(self):
+        return self._required_parameters
+
+    def get_values(self, parameter_values):
+        self._availible_values = set()
+        deny_values = set()
+
+        for e in self._values_conditions:
```
bugy, line 82: Could you rename `e` to something like `possible_value`, otherwise it's pretty hard to read

src/config/script/list_values.py
```diff
@@ -52,6 +52,46 @@ def __init__(self, script) -> None:
     def get_values(self, parameter_values):
         return self._values
 
+class DependantValuesProvider(ValuesProvider):
```
bugy, line 55: Would you mind adding a couple of tests for this class? Specifically for `get_values` method

bugy:
Hi @jtprof Thanks a lot for the PR! and really sorry for the late response, it was quite busy week for me.
I left a couple of small comments for the code.

Regarding python 3.6: I would love to use f-strings also, but I think there are still some people using python 3.5, so I would prefer to keep 3.5 compability for now, especially considering that 3.5 is still "active" python version https://devguide.python.org/#status-of-python-branches

src/config/constants.py
jtprof: Formally you are right. But 
1. Other types (such 'text', 'int', etc.) don't have its own constants, so the file needs to refactor
2. I believe the issue should been separated to its own type, so, it's for future use

jtprof:
> Hi @jtprof Thanks a lot for the PR! and really sorry for the late response, it was quite busy week for me.
No pressure.

> I left a couple of small comments for the code.
I will fix.

> 
> Regarding python 3.6: I would love to use f-strings also, but I think there are still some people using python 3.5, so I would prefer to keep 3.5 compability for now, especially considering that 3.5 is still "active" python version https://devguide.python.org/#status-of-python-branches

2020-09-13 'End-of-life'. It's so close.

.gitignore
```diff
@@ -76,4 +76,9 @@ web-src/selenium-debug.log
 web-src/chromedriver.log
 web-src/geckodriver.log
 
-venv/
\ No newline at end of file
+venv/
+
+conf/rediska.conf
```
bugy, line 81: I think those are not needed

src/config/constants.py
bugy: Can we call it "list subtype" or smth like this? :)
I raised this concern only because I was confused while reading the code

bugy:
HI @jtprof, sorry for being inactive here.

I started working on refactoring for this PR and thought how to make configuration of conditional values a bit simpler (at the moment it has multiple nested levels) and more generic (so that you can use patterns, comparison, etc.).
In the end, I came up with idea of writing simple conditions, like `"Dog food": "${param1} in ['a', 'b', 'c'] and ${param2} > 0"`
But then I realize, that this is already covered by a possibility to specify scripts. So instead of writing complicated config, you can always write a script and pass parameter values there. And then this script will be responsible for deciding which values are supported.
I understand, that for simple cases, which you try to cover here, a script might be overhead, but as for me, it still keeps config cleaner and reduces code complexity of script server

PR #321, opened 2020-07-13
Enable searchbox support on multiselect combobox

autogun:
* Add search box for `multiselect` type
* Position search box header to be on top of scrolled list
image


bugy:
Thanks!

PR #375, opened 2020-12-10
Max chars input validation

autogun:
Adding `max_chars` config parameter to a text field.

I sometime require limit user input to specific number of chars. I've been doing this until now with-in the executed script but I find it much useful this way.

### DIFF
--- a/src/model/parameter_config.py
+++ b/src/model/parameter_config.py
@@ -27,6 +27,7 @@ LOGGER = logging.getLogger('script_server.parameter_config')
     'type',
     'min',
     'max',
+    'max_length',
     'constant',
     '_values_provider',
     'values',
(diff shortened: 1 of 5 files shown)


src/model/parameter_config.py
```diff
@@ -277,6 +279,9 @@ def validate_value(self, value, *, ignore_required=False):
             return None
 
         if self.type == 'text':
+            if (not is_empty(self.max_chars)) and (len(value) > int(self.max_chars)):
+                return 'is greater than allowed char length (' \
```
bugy, line 283: Could you change "greater" to "longer?

src/model/parameter_config.py
```diff
@@ -27,6 +27,7 @@
     'type',
     'min',
     'max',
+    'max_chars',
```
bugy, line 30: may be we could call it "max_length"? What do you think?

autogun:
Got a little confused and mistakenly pushed something else. Please have another look. Apologies for that.

autogun:
@bugy is there anything else required here?

bugy:
Oh, damn, i forgot about it, sorry :(

bugy (approved):

PR #400, opened 2021-02-18
Honor X-headers for reverse proxy support

dnrce:
https://github.com/bugy/script-server/wiki/Reverse-Proxy-setup says to set the `X-Scheme` header but this is not respected because the corresponding flag is not set when instantiating `tornado.httpserver.HTTPServer`. This makes it impossible to terminate TLS on a reverse proxy.

### DIFF
--- a/src/tests/web/server_test.py
+++ b/src/tests/web/server_test.py
@@ -88,6 +88,14 @@ class ServerTest(TestCase):
             {'name': 's3', 'group': None}],
             response['scripts'])
 
+    def test_redirect_honors_protocol_header(self):
+        self.start_server(12345, '127.0.0.1')
+
+        response = requests.get('http://127.0.0.1:12345/',
+            allow_redirects=False,
+            headers={'X-Forwarded-Proto': 'https'})
+        self.assertRegex(response.headers['Location'], '^https')
+
     def request(self, method, url):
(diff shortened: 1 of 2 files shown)


bugy:
Cool, thanks a lot!

bugy (approved):

bugy:
Hi @dnrce, unfortunately setting xheaders causing problems with `trusted_ips` setting (https://github.com/bugy/script-server/issues/405#issuecomment-791903109). Also, I think, that enabling xheaders for IPs by default might be a security risk.
That's why I'd like to disable xheaders back and implement a custom handling for X-Scheme (I guess it's fine to always enable X-Scheme). 

However, I cannot reproduce your original issue with TLS connection, maybe you could advise me on how to do it? I checked https://script-server.net/index.html and it works fine for me, I don't see any errors in browser or server logs.

Nottt:
Let me know when I can test it again please

dnrce:
Yeah, you're right on both fronts, unfortunately. I don't think I have time to work on a more complete solution but I'll file this as an open issue with reproduction steps.

bugy:
No problem, I'll work on it. Just need reproduction steps

dnrce:
Thanks! Filed #406 for this.

PR #415, opened 2021-03-24
Allow json comments

brunomgalmeida:
This may be useful, to document scripts or features


bugy:
Hi @brunomgalmeida I think this is a helpful feature.
However, I prefer to avoid adding extra dependencies without extra need for several reasons:
1. It breaks backward compatibility somehow (after installing new version, the server won't start)
2. Some dependencies might introduce security risk or just some bugs, and being not very popular, the risk is higher
3. I would like to keep machines as lightweight as possible

samples/conf.json
```diff
@@ -0,0 +1,8 @@
+{
+  "title": "Script server",
+  "port": "8000"
+//   "ssl": {
```
bugy, line 4: Could you add some normal comment instead, please? 
And keep the sample configuration

src/migrations/migrate.py
```diff
@@ -207,7 +208,7 @@ def __introduce_access_config(context):
         return
 
     content = file_utils.read_file(file_path)
-    json_object = json.loads(content, object_pairs_hook=OrderedDict)
+    json_object = custom_json.loads(content, object_pairs_hook=OrderedDict)
```
bugy, line 211: Could you also update "_load_runner_files" ?

src/utils/custom_json.py
```diff
@@ -0,0 +1,12 @@
+import json
+import re
+
+def loads (content):
+    contents=""
+    for line in content.split('\n'):
+        if not re.match (r'\s*//.*', line):
+            contents += line + "\n"
+    # while "/*" in contents:
```
bugy, line 9: Could you remove the commented out code, please?

src/utils/custom_json.py
```diff
@@ -0,0 +1,12 @@
+import json
+import re
+
+def loads (content):
+    contents=""
```
bugy, line 5: Could you replace double-quotes with single quotes?

src/model/server_conf.py
```diff
@@ -62,7 +63,7 @@ def from_json(conf_path, temp_folder):
 
     config = ServerConfig()
 
-    json_object = json.loads(file_content)
+    json_object = custom_json.loads(file_content)
```
bugy, line 66: Could you add a test for it, please?

src/config/config_service.py
```diff
@@ -108,7 +109,7 @@ def list_configs(self, user, mode=None):
 
         def load_script(path, content):
             try:
-                json_object = json.loads(content)
+                json_object = custom_json.loads(content)
```
bugy, line 112: Could you also add a test for loading script configs with comments, please?

bugy (changes requested):
Thanks for the changes! I left a couple of comments

Also, I think these places should be updated as well:
- auth_ldap:255
- config_service:192
- script_config:246
- schedule_service:38

samples/conf.json
brunomgalmeida: done

bugy (approved):
Thank you!

PR #423, opened 2021-03-30
Mutiple include

brunomgalmeida:
Allow import of multiple files
will only care about the json section "parameters"


bugy (changes requested):
Hey, I think the idea of the PR is super nice and thanks for the initial changes!
Unfotunately, I think we should improve the implementation. I left a couple of comments in the code. Additionally, we would need to consider the following things:
- Each included path can depend on other parameters: e.g. `"include": "configs/${list1}.json"` That's why there was "_TemplateProperty" used. Unfortunately for the list this would be much more difficult to implement (we would need to listen on changes in any element of the list)
- Could you add a couple of tests for it?

bugy (approved):

muzzol:
sorry to resurface this old issue.

can this solution be used in conf.json?

for security purposes I need a separated file with access info but I don't see any include example on main conf file.

is this possible?

bugy:
Hey @muzzol unfortunately not :(

Issue #383, opened 2020-12-22
Ability to sort History columns


jarrah31:
Would it be possible to have the ability to sort History columns alphabetically/numerically please? e.g. if I click on the Script column header, it sorts the list A-Z, and clicking it again sorts it Z-A. 

Thanks!

bugy:
Implemented by @autogun :+1:

jarrah31:
Thank you!

PR #430, opened 2021-04-13
Ability to sort History columns

autogun:
This commit adds the ability to sort history columns, I tried to keep it lean and simple.
Issue https://github.com/bugy/script-server/issues/383

### DIFF
--- a/web-src/src/common/components/history/executions-log-table.vue
+++ b/web-src/src/common/components/history/executions-log-table.vue
@@ -79,6 +79,10 @@ export default {
 </script>
 
 <style scoped>
+.executions-log-table th  {
+  cursor: pointer;
+}
+
 .executions-log-table tbody > tr {
   cursor: pointer;
 }
--- a/web-src/src/common/components/history/executions-log-table.vue
+++ b/web-src/src/common/components/history/executions-log-table.vue
@@ -3,11 +3,11 @@
     <table class="highlight striped">
       <thead>
       <tr>
-        <th class="id-column" :class="showArrow('id')" @click="sortBy('id')">ID</th>
-        <th class="start_time-column" :class="showArrow('startTimeString')" @click="sortBy('startTimeString')">Start Time</th>
-        <th class="user-column" :class="showArrow('user')" @click="sortBy('user')">User</th>
-        <th class="script-column" :class="showArrow('script')" @click="sortBy('script')">Script</th>
-        <th class="status-column" :class="showArrow('fullStatus')" @click="sortBy('fullStatus')">Status</th>
+        <th class="id-column" :class="showSort('id')" @click="sortBy('id')">ID</th>
+        <th class="start_time-column" :class="showSort('startTimeString')" @click="sortBy('startTimeString')">Start Time</th>
+        <th class="user-column" :class="showSort('user')" @click="sortBy('user')">User</th>
+        <th class="script-column" :class="showSort('script')" @click="sortBy('script')">Script</th>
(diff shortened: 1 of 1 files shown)


bugy:
Hi, @autogun thanks a lot for your changes!
Could you also add arrow showing currently sorted column and the direction? 

I think this is a nice example, doing it using css:
https://codepen.io/marcmatias/pen/gxPzvY

Search for "arrow_back" in css section: the example code snippet adds an `:after` pseudoelement, rendering the arrow
(the Material Icons font should be available already)

autogun:
Hi @bugy, please see below -
https://user-images.githubusercontent.com/11921235/114769435-e3e1f100-9d72-11eb-9a50-2497afe61f64.mov

web-src/src/common/components/history/executions-log-table.vue
```diff
@@ -3,11 +3,21 @@
     <table class="highlight striped">
       <thead>
       <tr>
-        <th class="id-column">ID</th>
-        <th class="start_time-column">Start Time</th>
-        <th class="user-column">User</th>
-        <th class="script-column">Script</th>
-        <th class="status-column">Status</th>
+        <th class="id-column">
+          <a href="#" @click="sortBy('id')">ID<span :class="ascending ? 'arrow asc' : 'arrow desc'"></span></a>
```
bugy, line 7: Instead of `arrow asc` I would suggest using just `asc`. The same for `desc`.

web-src/src/common/components/history/executions-log-table.vue
```diff
@@ -72,4 +113,25 @@ export default {
...
+.executions-log-table .arrow.asc {
+  border-left: 4px solid transparent;
+  border-right: 4px solid transparent;
+  border-bottom: 4px solid #42b983;
+}
+
+.executions-log-table .arrow.desc {
+  border-left: 4px solid transparent;
+  border-right: 4px solid transparent;
+  border-top: 4px solid #42b983;
```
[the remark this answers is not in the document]
bugy: Please also set `color: var(--font-color-main)` for the anchor element, so it looks as before:
image
(the table header is not really a link, so it shouldn't look like one, in my opinion)

Or may be you could even remove the `<a>` elements and put onclick handler directly to `<th>`. The only adjustment which would be needed, is to set css cursor:pointer for it

bugy (changes requested):
Hey, I left a couple of improvement suggestions.
Also, I think we should show arrows only for the sorted column (not for every) :sweat_smile:

autogun:
Hope I got it right this time 😄 

https://user-images.githubusercontent.com/11921235/114936275-e280f900-9e44-11eb-97e6-6d73a8f6361e.mov

web-src/src/common/components/history/executions-log-table.vue
autogun: How should the `css` part look like if I drop `arrow`?

bugy: Instead of `.arrow.asc` you could use `th.asc` (tag name)

bugy: Ah, I got what you mean. Probably you could just rename "arrow" to "sorted" ?

autogun: Sure thing, done.

bugy:
Looks much better now, thanks! :) 
The last tiny thing: could you add css `cursor: pointer` for `th` tag? So that when you mouse-over, the cursor changes

bugy:
Thanks s lot!

bugy (approved):

Issue #442, opened 2021-05-23
Load data from script execution doesn't work


drehelis:
Hi,

Seems like below feature is broken on `dev`
https://github.com/bugy/script-server/wiki/Script-config#script

```
  "parameters": [
    {
      "name": "Test",
      "param": "--profile",
      "type": "list",
      "values": [
        "Test 1",
        "Test 2"
      ],
      "required": true
    },
    {
      "name": "Test 2",
      "param": "--test",
      "values": {
        "script": "echo ${Test}"
      },
      "required": true
    }
  ]
```

This used to work in `1.6`

bugy:
Hi @autogun it works for me, (if I add `"type": "list"` to the second parameter)

bugy:
![image](https://user-images.githubusercontent.com/1275813/119256758-86875e00-bbc2-11eb-904e-3a11d8e7c60f.png)

autogun:
I can swear it also worked with `"type": "text"`, no?

bugy:
I don't think so. How is it supposed to work?

autogun:
This is embarrassing, I think I have custom code which I forgot about :-D

Issue #320, opened 2020-07-12
default value and variable error


antonellocaroli:
Hi,
I'm not sure if this is expected behavior or a bug.

###############################
{
      "name": "Select seftware to set",
      "param": "-b",
      "type": "list",
      "default": "networkaudiod",
      "values": [
        "networkaudiod",
        "squeezelite-R2",
        "squeezelite"
      ],
      "description": "Select seftware to set"
    },
    {
      "name": "Set affinity",
      "param": "-c",
      "type": "text",
      "default": {
        "script": "/opt/.gentooplayer/web/scripts/script/expert1/${Select seftware to set}/affinity.sh"
      },
      "description": "Set CPU affinity"
    },
################################

2020-07-12 07:27:23,794 [root.INFO] Starting GentooPlayer, v1.16.0-master@995fec1
2020-07-12 07:27:23,811 [asyncio.DEBUG] Using selector: EpollSelector
Server is running on: http://0.0.0.0:5000
2020-07-12 07:27:27,678 [identification.INFO] Assigned user_id=192.168.178.22-6d9fb38bcbd547e6 to {'ip': '192.168.178.22', 'hostname': 'GentooServer.fritz.box'}
2020-07-12 07:27:27,719 [tornado.access.INFO] 304 GET /index.html (192.168.178.22) 45.99ms
2020-07-12 07:27:27,748 [tornado.access.INFO] 304 GET /css/chunk-index-vendors.47bd9eac.css (192.168.178.22) 3.77ms
2020-07-12 07:27:27,758 [tornado.access.INFO] 304 GET /css/index.00b840dd.css (192.168.178.22) 7.60ms
2020-07-12 07:27:27,763 [tornado.access.INFO] 304 GET /js/chunk-index-vendors.3dee9c44.js (192.168.178.22) 10.09ms
2020-07-12 07:27:27,766 [tornado.access.INFO] 304 GET /js/index.807c7662.js (192.168.178.22) 11.71ms
2020-07-12 07:27:27,800 [tornado.access.INFO] 304 GET /favicon.ico (192.168.178.22) 3.14ms
2020-07-12 07:27:27,836 [tornado.access.INFO] 304 GET /img/titleBackground_small.8d99a14e.jpg (192.168.178.22) 3.40ms
2020-07-12 07:27:27,842 [tornado.access.INFO] 304 GET /executions/active (192.168.178.22) 3.13ms
[...]

Issue #141, opened 2018-11-26
Add possibility to specify a result of a command as value (the same thing than for list values)


csphoenix1:
Hi !

Is it possible to add the possibility to specify a result of a command as the value of a parameter ?

Like that but not only for a list values.

`{
      "name": "Reload the current profile",
      "value": {
        "script": "cat /tmp/current_profile.txt"
      }
}`

bugy:
Hi @csphoenix1, do you mean for constants and default values?

csphoenix1:
Hi @bugy 

Yes for constant and default value, exactly.

bugy:
Done. Usage example:
```
      "default": {
        "script": "shuf -i 10000-99999 -n 1"
      }
```

PR #445, opened 2021-05-25
Textfield to support dependant values

drehelis:
Following my embarrassing issue https://github.com/bugy/script-server/issues/442, trying to cherry-pick this change to hopefully merge it to master.

I'm not entirely sure the implementation is right thought. Meanwhile, this is what I got -

https://user-images.githubusercontent.com/11921235/119562769-65379500-bdaf-11eb-981b-f730d291ad82.mov


autogun:
Yes, `Text-A` is updated accordingly to the `Drop-A` selection.

By use `default`, you mean from the configuration point of view? 
To change this:
```
    {
      "name": "Text-A",
      "values": {
        "script": "echo ${Drop-A}_"
      },
      "required": true
    },
```
to:
```
    {
      "name": "Text-A",
      "default": {
        "script": "echo ${Drop-A}_"
      },
      "required": true
    },
```

bugy:
Yes, exactly this one

autogun:
How about consistency? Across all config, `default` is `key:value` and not an object.
While values are all objects with parameter injection ability.

bugy:
Apparently, `script` option is already supported there: #141 :sweat_smile: 
But it doesn't support dependant parameters yet

As for me, the word `default` reflects the behaviour better, than `values`

autogun:
@bugy, one way to achieve this is in my last commit, but that still uses the infrastructure of `values`.
This, however, allows me to use both `default` and `values` in a text config.

bugy:
Hi @autogun, sorry for taking so long. To be honest, I would prefer to keep the code explicit, in this sense I would suggest to introduce explicit `default_value_provider`, which would set a single default value. 
Then you would need to extend `model.parameter_config.ParameterModel._param_values_observer` to work with the default values provider
And create a method similar to `model.parameter_config.ParameterModel._reload_values`, which would set `default`

On the client side, you would need to listen to "default" changes (not values)
When an old default value still equals to `this.value`, then you can update it, by calling the following code:
```
      this._doValidation(newValue);
      this.$emit('input', newValue);
```

bugy:
supported via "default" option in scope of #320. So closing this one as obsolete.

PR #457, opened 2021-06-08
Correctly sort History columns

autogun:
Sorting history columns introduced in https://github.com/bugy/script-server/pull/430 does not take into account that by default, `sort()` method sorts elements alphabetically. This leads to an unwanted sort result of: `[1, 10, 100, 2, 3]`

This PR enhances sorting to sort numerically, by date property etc.

### DIFF
--- a/samples/scripts/test.sh
+++ b/samples/scripts/test.sh
@@ -1,0 +1,3 @@
+#!/bin/bash
+
+echo $@
--- a/samples/scripts/test.sh
+++ b/samples/scripts/test.sh
@@ -1,3 +1,0 @@
-#!/bin/bash
-
-echo $@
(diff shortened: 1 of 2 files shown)


samples/scripts/test.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
```
bugy, line 1: Is this file needed?

bugy (changes requested):
One small comment, otherwise looks good

samples/scripts/test.sh
autogun: Definitely not, slipped in by mistake :-)

bugy (approved):
Thanks!

bugy:
Hi, i found another issue: script name sorting considers letter case, could
you fix it also, please? :)

Danny Rehelis ***@***.***> schrieb am Mi., 9. Juni 2021,
12:46:

bugy (approved):

bugy:
Thanks again!

Issue #505, opened 2021-11-19
Scheduling less than hour and checkbox on admin


sarakha63:
Hello i have a minor easy request about scheduling.

It could be great to have a checkbox to enable scheduling on admin page

And for scheduling it would be great to have minutes also

As of now i have some scritp on script server machine that i add to crontab to make them do a health check on some services each 10 minutes (example) and i do that on crontab  and i would love to handle all scripting processes trough script server.

Both of those requests are i think simple quick features to implement. If you could consider it.

By the way thanks for this webapp, i can say after a lot of digging that nothing like that exists that is simple enough to give access for my wife for example

PR #526, opened 2022-02-24
Add the ability to schedule a script by minutes

### LINKED ISSUE #458: Schedule increments
Hi - it would be great to have a schedule running every 30 minutes. From looking, the minimum is every 1 hour. Can this be changed to allow us to run scripts more frequently? Thanks.

Conway:
This adds the ability to schedule a script to run on every x minutes, instead of 1 hour being the minimum interval

Addresses #505, closes #458


bugy (approved):
Thank you

PR #571, opened 2022-08-30
Add YAML runner config support

anthony-francisco:


src/utils/custom_yaml.py
```diff
@@ -0,0 +1,10 @@
+import json
+import re
+import yaml
+
+def loads (content, **args):
+    contents=''
+    for line in content.split('\n'):
+        if not re.match (r'\s*//.*', line):
+            contents += line + "\n"
```
bugy, line 9: I believe this is not needed. Yaml supports comments by default (in the specification and in the library)

src/config/config_service.py
```diff
@@ -231,7 +234,10 @@ def _visit_script_configs(self, visitor):
     def _find_config(self, name) -> Union[ConfigSearchResult, None]:
         def find_and_load(path, content):
             try:
-                json_object = custom_json.loads(content)
+                if path.endswith('.yaml'):
+                    json_object = custom_yaml.loads(content)
+                else:
+                    json_object = custom_json.loads(content)
```
bugy, line 240: This is a same code as in line 173. Could you extract it to a common method?

src/utils/custom_yaml.py
```diff
@@ -0,0 +1,10 @@
+import json
+import re
+import yaml
```
bugy, line 3: This is a non-standard library and since most of script server users don't need it, the module should be optional.
In order to make it work for others, the import statement should be moved inside `loads` function. I.e. it will be imported only when we try to read a yaml file.

bugy (changes requested):

bugy:
Hi @anthony-francisco thanks for PR! I left a couple of comments which would be cool to solve

anthony-francisco:
Hi @bugy , I made some changes (regarding your comments).
I hope it will be now OK for you :)

bugy:
Hi @anthony-francisco thanks :) Could you also check my comment in custom_yaml.py?

bugy (approved):

bugy:
Great, thank you for the work!

PR #577, opened 2022-09-21
Feature: history search#572

gpsinghsandhu:
A very simple implementation of search functionality in history. Search is implemented client-side. This currently searches only on 'user' and 'script' for the given search text. Suggestions are welcome.

Also fixed vue warning regarding mutating props. See commit - cdf8352891251e448b1858eaab0de8a62d608476

@bugy Need input on where to add this in documentation.

Note: I did not use the current 'SearchPanel' component, since it seemed very specific to the script search.
Screenshot 2022-10-06 at 12 28 24 PM
Screenshot 2022-10-06 at 12 28 36 PM

### DIFF
--- a/web-src/src/common/components/history/executions-log-table.vue
+++ b/web-src/src/common/components/history/executions-log-table.vue
@@ -200,28 +200,9 @@ export default {
 
 .search-panel {
   display: flex;
-  padding: 5px;
-  border:  1px solid var(--primary-color);
-  border-radius: 4px;
-  background-color: var(--background-color-high-emphasis);
 }
 
 .search-button {
   align-self: center;
-}
-
-input.search-field {
-  height: 1.5rem;
-  font-size: 1rem;
-  float: right;
-  padding: 0;
-  margin: 0;
-  border: 0;
-  box-shadow: none;
-}
-
-input.search-field:not([type]):focus {
-  border-bottom: 0;
-  box-shadow: none;
 }
 </style>
--- a/web-src/src/common/components/history/executions-log-table.vue
+++ b/web-src/src/common/components/history/executions-log-table.vue
@@ -38,25 +38,13 @@
 
 <script>
 import {mapState} from 'vuex';
-import SearchPanel from '../../../main-app/components/SearchPanel'
 import ClearIcon from '@/assets/clear.png'
 import SearchIcon from '@/assets/search.png'
 
 export default {
   name: 'executions-log-table',
-  components: {
-    SearchPanel
-  },
   props: {
     rows: Array,
-    'sortColumn': {
-      type: String,
-      default: 'id'
-    },
-    'ascending': {
-      type: Boolean,
-      default: false
-    },
     rowClick: {
       type: Function
     }
@@ -65,24 +53,24 @@ export default {
   data() {
     return {
       searchText: '',
-      mySortColumn: this.sortColumn,
-      myAscending: this.ascending
+      sortColumn: 'id',
+      ascending: false
     }
   },
 
   methods: {
     showSort: function (sortKey) {
-      if (this.mySortColumn === sortKey) {
-        return this.myAscending ? 'sorted asc' : 'sorted desc'
+      if (this.sortColumn === sortKey) {
+        return this.ascending ? 'sorted asc' : 'sorted desc'
       }
     },
 
     sortBy: function (sortKey) {
-      if (this.mySortColumn === sortKey) {
-        this.myAscending = !this.myAscending;
+      if (this.sortColumn === sortKey) {
+        this.ascending = !this.ascending;
       } else {
-        this.myAscending = true;
-        this.mySortColumn = sortKey;
+        this.ascending = true;
+        this.sortColumn = sortKey;
       }
     },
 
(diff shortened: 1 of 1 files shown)


web-src/src/common/components/history/executions-log-table.vue
```diff
@@ -44,38 +62,64 @@ export default {
     }
   },
 
+  data() {
+    return {
+      filteredRows: this.rows ? [...this.rows] : [],
```
bugy, line 67: Instead of using data+watch, it's better to use computed property.

web-src/src/common/components/history/executions-log-table.vue
```diff
@@ -44,38 +62,64 @@ export default {
...
+      searchText: '',
+      mySortColumn: this.sortColumn,
+      myAscending: this.ascending
+    }
+  },
+
+  watch: {
+    rows: function(val, oldVal) {
+      this.searchText = '';
```
bugy, line 76: I think this is not needed. If new data is loaded, I don't see any reason to reset search string.

web-src/src/common/components/history/executions-log-table.vue
```diff
@@ -44,38 +62,64 @@ export default {
     }
   },
 
+  data() {
+    return {
+      filteredRows: this.rows ? [...this.rows] : [],
+      searchText: '',
+      mySortColumn: this.sortColumn,
+      myAscending: this.ascending
```
bugy, line 70: Why do we need those 2? I think properties should be enough

bugy (changes requested):
Hi @gpsinghsandhu thanks for your PR! And sorry for super slow response :(
In general, the idea is good, but I would like to have it to be more Vue/reactive style, with having computed properties, instead of data+watch.

web-src/src/common/components/history/executions-log-table.vue
gpsinghsandhu: Got it. Just very new to vue, so learning.

web-src/src/common/components/history/executions-log-table.vue
gpsinghsandhu: Sure. This behaviour is also fine.

web-src/src/common/components/history/executions-log-table.vue
gpsinghsandhu: The reason for this change is - since we are changing the value of these properties, vue is giving warnings. Attaching the warning screen shots
Screenshot 2022-09-21 at 6 18 44 PM
Screenshot 2022-09-21 at 6 18 52 PM

web-src/src/common/components/history/executions-log-table.vue
```diff
@@ -26,9 +38,15 @@
...
 import {mapState} from 'vuex';
+import SearchPanel from '../../../main-app/components/SearchPanel'
+import ClearIcon from '@/assets/clear.png'
+import SearchIcon from '@/assets/search.png'
 
 export default {
   name: 'executions-log-table',
+  components: {
+    SearchPanel
+  },
```
bugy, line 49: Please remove it, this is not needed.

web-src/src/common/components/history/executions-log-table.vue
bugy: Oh, I see, thanks. I think it was a mistake in the previous PR. ascending and sortColumn should be defined as data, not props. Could you remove them from props, please and just use `ascending` and `sortColumn` names in data

bugy (changes requested):

web-src/src/common/components/history/executions-log-table.vue
```diff
@@ -149,4 +192,36 @@ export default {
...
+  padding: 0;
+  margin: 0;
+  border: 0;
+  box-shadow: none;
+}
+
+input.search-field:not([type]):focus {
+  border-bottom: 0;
+  box-shadow: none;
+}
```
bugy, line 226:
```suggestion
.search-panel {
  display: flex;
}

.search-button {
  align-self: center;
}
```

let's keep existing script server styling and don't introduce new styles, it makes the whole project look as a whole and cleaner :)

web-src/src/common/components/history/executions-log-table.vue
```diff
@@ -58,24 +72,57 @@ export default {
         this.ascending = true;
         this.sortColumn = sortKey;
       }
+    },
+
+    searchIconClickHandler() {
+      if (this.isClearSearchButton) {
```
bugy, line 78: I would suggest to check `searchText` instead

Also, if searchText is empty, probably we could focus the field, when the button is clicked

bugy (changes requested):
Hi, I finally found some time and checked the behaviour locally, it would be great to improve the button behaviour and use default materialize styles for the input field :)

web-src/src/common/components/history/executions-log-table.vue
gpsinghsandhu: Done.

I believe we should focus the field in any case (whether clearing or clicking search button).

bugy (approved):
Thanks!

Issue #599, opened 2022-11-27
Pass execution id to a script


bugy:
Reported by @matiasrivera:
>  Is it somehow possible to get the running ID from the inside the running script? [...] I would love to know if it is possible to receive the running ID or read it from somewhere.
> I have a need to somehow allow my users to identify their runs from some of the parameters they choose or the script generated. [...] I have all these simultaneous runs of the same script from which I'd love to know some of the parameters I used to execute them from the history pane. Is it possible to expose some kind of information?

bugy:
I added it via environment variable, EXECUTION_ID. Unfortunately passing it as a parameter is more complicated

PR #618, opened 2023-01-10
add current audit informations in env variables to running scripts

seb0uil:
Simply expose audit info in env variables to running script.

I keep the #599 env variables pattern and expose audit info with _EXECUTION_AUDIT__ variables


bugy:
as discussed in a ticket, this is already possible to do via parameter. So I'm closing this PR, as I don't want to have different ways to achieve the same functionality.

PR #643, opened 2023-03-31
Added delete script on frontend

EduardSchwarzkopf:
Please provide any feedback. I'm not a front-end guy and never touched that much Vue code before.

Thanks

### DIFF
--- a/src/config/config_service.py
+++ b/src/config/config_service.py
@@ -49,10 +49,10 @@ def _preprocess_incoming_config(config):
     config['name'] = name.strip()
 
 
-def _add_date_prefix_to_filename(filename):
+def _create_archive_filename(filename):
     current_datetime = datetime.now()
     formatted_datetime = current_datetime.strftime('%Y%m%d%H%M%S')
-    return f"{formatted_datetime}_{filename}"
+    return f"{formatted_datetime}_{filename}.deleted"
 
 
 class ConfigService:
@@ -148,11 +148,8 @@ class ConfigService:
                 f'{str(user)} has no admin access to {short_config.name}'
             )
 
-        archived_configs_folder = os.path.join(self._script_configs_folder, '..', 'deleted')
-        file_utils.prepare_folder(archived_configs_folder)
-
-        archive_file_name = _add_date_prefix_to_filename(os.path.basename(path))
-        archive_file_path = os.path.join(archived_configs_folder, archive_file_name)
+        archive_file_name = _create_archive_filename(os.path.basename(path))
+        archive_file_path = os.path.join(self._scripts_deleted_folder, archive_file_name)
         unique_archive_file_path = file_utils.create_unique_filename(archive_file_path, 100)
 
         LOGGER.info(
--- a/src/config/config_service.py
+++ b/src/config/config_service.py
@@ -60,9 +60,11 @@ class ConfigService:
         self._authorizer = authorizer  # type: Authorizer
         self._script_configs_folder = os.path.join(conf_folder, 'runners')
         self._scripts_folder = os.path.join(conf_folder, 'scripts')
+        self._scripts_deleted_folder = os.path.join(conf_folder, 'deleted')
         self._process_invoker = process_invoker
 
         file_utils.prepare_folder(self._script_configs_folder)
+        file_utils.prepare_folder(self._scripts_deleted_folder)
 
     def load_config(self, name, user):
         self._check_admin_access(user)
--- a/src/config/config_service.py
+++ b/src/config/config_service.py
@@ -52,7 +52,7 @@ def _preprocess_incoming_config(config):
 def _create_archive_filename(filename):
     current_datetime = datetime.now()
     formatted_datetime = current_datetime.strftime('%Y%m%d%H%M%S')
-    return f"{formatted_datetime}_{filename}.deleted"
+    return f"{formatted.datetime}_{filename}.deleted"
 
 
 class ConfigService:
--- a/src/config/config_service.py
+++ b/src/config/config_service.py
@@ -2,6 +2,7 @@ import json
 import logging
 import os
 import re
+import shutil
 from typing import NamedTuple, Optional
 
 from auth.authorization import Authorizer
@@ -13,6 +14,8 @@ from utils import os_utils, file_utils, process_utils, custom_json, custom_yaml
 from utils.file_utils import to_filename
 from utils.process_utils import ProcessInvoker
 from utils.string_utils import is_blank, strip
+from datetime import datetime
+
 
 SCRIPT_EDIT_CODE_MODE = 'new_code'
 SCRIPT_EDIT_UPLOAD_MODE = 'upload_script'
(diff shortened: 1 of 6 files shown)


src/config/config_service.py
```diff
@@ -122,6 +131,33 @@ def update_config(self, user, config, filename, uploaded_script):
...
+
+        (short_config, path, config_object) = search_result
+
+        if not self._can_edit_script(user, short_config):
+            raise ConfigNotAllowedException(
+                f'{str(user)} has no admin access to {short_config.name}'
+            )
+
+        archived_configs_folder = os.path.join(self._script_configs_folder, '..', 'deleted')
+        file_utils.prepare_folder(archived_configs_folder)
```
bugy, line 150: May be we could initialize this in constructor?

src/web/server.py
```diff
@@ -202,6 +202,27 @@ def get(self, user, script_name):
...
+            respond_error(self, 403, 'Access to the script is denied')
+            return
+        except NotFoundException as e:
+            LOGGER.warning(f'Failed to delete script {script_name}', exc_info=True)
+            respond_error(self, 404, str(e))
+            return
+        except InvalidAccessException as e:
+            raise tornado.web.HTTPError(403, reason=str(e)) from e
+
+        self.set_status(204)
```
bugy, line 223: I would remove this line. Currently, script server always responds with 200 in all the endpoints.
And to be honest, I don't see any benefits of success http codes.

web-src/src/admin/components/scripts-config/ScriptConfig.vue
```diff
@@ -33,14 +37,20 @@ export default {
   },
 
   methods: {
-    ...mapActions('scriptConfig', ['init', 'save'])
+    ...mapActions('scriptConfig', ['init', 'save', 'deleteScript']),
+    async onDeleteScript() {
+        await this.deleteScript({scriptName: this.scriptName});
+    }
```
bugy, line 43: I think this is redundant, you can just call `deleteScript` from `<PromisableButton :click="onDeleteScript"`

web-src/src/admin/store/script-config-module.js
```diff
@@ -80,7 +80,21 @@ export default {
                     }
                     throw e;
                 });
-        }
+        },
+
+        deleteScript({}, {scriptName}) {
+            return axiosInstance.delete(`admin/scripts/${encodeURIComponent(scriptName)}`).then(() => {
```
bugy, line 86: could you move `.then` on a new line please? I think starting every new operation on a new line makes code a bit easier to read

src/config/config_service.py
```diff
@@ -122,6 +131,33 @@ def update_config(self, user, config, filename, uploaded_script):
...
+
+        archive_file_name = _add_date_prefix_to_filename(os.path.basename(path))
```
bugy, line 152: You should also change an extension to be different from `json`, otherwise script server will find this configuration.
I think it could be either `._json` or `.json.deleted`

web-src/src/admin/components/scripts-config/ScriptConfig.vue
```diff
@@ -8,6 +8,9 @@
           <h5>Parameters</h5>
           <ScriptParamList :parameters="scriptConfig.parameters"/>
         </div>
+        <div v-if="scriptName !== NEW_SCRIPT">
```
bugy, line 11: Should we put it in the footer? Where Save button resides

bugy (changes requested):
Hi, in general looks good to me, thanks!
There are just some minor improvement suggestions

src/web/server.py
EduardSchwarzkopf: Yeah, I'm fine with that. Less code is good code.

web-src/src/admin/components/scripts-config/ScriptConfig.vue
EduardSchwarzkopf: when i did that without the await it threw me an error in the console `TypeError: this.click() is undefined`. I could not figure out why, but this way it is not showing up, when the button gets clicked

web-src/src/admin/store/script-config-module.js
EduardSchwarzkopf: yes sure. Are there maybe even some ESLinter options defined? Then I could simply let Prettier handle all the formatting.

src/config/config_service.py
EduardSchwarzkopf: ah yes you are right, i forgot about this

web-src/src/admin/components/scripts-config/ScriptConfig.vue
EduardSchwarzkopf: It might be more fitting, because the rest belongs to the script, and the save/create/delete are just for storing them

src/config/config_service.py
EduardSchwarzkopf: Yes, this even might come in handy for a later function where you want to restore the scripts from the frontend.

web-src/src/admin/components/scripts-config/ScriptConfig.vue
EduardSchwarzkopf: hmm... I'm not sure to style this properly in this case 😅
image

bugy: I pushed a commit with moving and styling this button
FYI I removed the background, to avoid extra attention to this button on UI + in order to avoid having static colors (because they don't play nicely with dark theme support)

src/config/config_service.py
```diff
@@ -46,14 +49,22 @@ def _preprocess_incoming_config(config):
     config['name'] = name.strip()
 
 
+def _create_archive_filename(filename):
+    current_datetime = datetime.now()
+    formatted_datetime = current_datetime.strftime('%Y%m%d%H%M%S')
+    return f"{formatted_datetime}_{filename}.deleted"
```
bugy, line 55: Should we use `.` instead of `_` for date separator? With underscore it looks a bit like a part of an original name (at least for me, because I use `_` for files naming, e.g. `bash_formatting.json`)

src/web/server.py
```diff
@@ -201,6 +201,26 @@ def get(self, user, script_name):
...
+            respond_error(self, 403, 'Access to the script is denied')
+            return
+        except NotFoundException as e:
+            LOGGER.warning(f'Failed to delete script {script_name}', exc_info=True)
+            respond_error(self, 404, str(e))
+            return
+        except InvalidAccessException as e:
+            raise tornado.web.HTTPError(403, reason=str(e)) from e
+
+        return
```
bugy, line 222: This return is redundant

web-src/src/admin/components/scripts-config/ScriptConfig.vue
bugy: To be honest I don't know how you did it.
I just used the same approach as in "save" button and it worked for me. I pushed this change to your PR

bugy (approved):
Looks good to me. I added 2 commits to address my previous comments, feel free to check them.

And let me know, if this PR is ready to be merged or you would like to add/test something else :)

web-src/src/admin/components/scripts-config/ScriptConfig.vue
EduardSchwarzkopf: Nicely done. If it is doing it's thing without any errors, that's a win

EduardSchwarzkopf:
Looking great! Thanks for giving me feedback on this.

src/config/config_service.py
EduardSchwarzkopf: yeah sure that can be a `.`

bugy:
Merged, thanks a lot @EduardSchwarzkopf for the contribution!

Issue #676, opened 2023-06-09
Schedule repeat end date / execution count


UrekD:
Hey,

For scheduling repeatable task it would be nice to have extra options to specify when the task should stop/not be executed anymore or how many times a task should execute.

For I instance I reworked the current repeatable option to require the end time to satisfy our needs as:
![image](https://github.com/bugy/script-server/assets/38784343/5b27f1b3-2e8b-41cc-b8b9-515e7a9ace8e)

I'm open to working on a proper implementation, but I'd need someone to do the frontend, as I'm terrible with that.

bugy:
Hi @UrekD great idea, thanks. Feel free to submit PR for backend. 
You can try to do frontend too, it should be pretty simple in this case. I could tell you which classes are affected

UrekD:
> Hi @UrekD great idea, thanks. Feel free to submit PR for backend. You can try to do frontend too, it should be pretty simple in this case. I could tell you which classes are affected

Hey @bugy, yeah I figured it would be much easier to edit the vue and build it than the manual js editing I did like an idiot before xD

After managing to get it to build on windows, I managed to construct the box shown below.

![brave_dpBB3Fl6x2](https://github.com/bugy/script-server/assets/38784343/7e5eea0d-225b-42b6-8983-9985f43cd258)

If you are happy with the structure of that I'll move onto the functional part. As you can see the styling is not good with that, if there is an easy way to handle that I'd appreciate that. The yet unformatted additions can be found [here](https://github.com/bugy/script-server/compare/master...UrekD:script-server:master).

bugy:
Looks very good to me! If you struggle too much with styling, you can leave it for me

UrekD:
Minor bug the date field is not required, so user can sent it empty as it just grabs the default value.

![image](https://github.com/bugy/script-server/assets/38784343/10ad6f37-d04a-413b-ba1f-ae7f63700caf)

UrekD:
> Looks very good to me! If you struggle too much with styling, you can leave it for me

Ah I thought the styling was dynamic not static, in this case I got it.

![brave_XAUBlsHBCw](https://github.com/bugy/script-server/assets/38784343/a6fc127b-0e2b-42bd-9d76-3be1d83c7cb7)

I think I got the functionality down I'll give it a day or 2 test then open merge 😄

PR #677, opened 2023-06-14
Extend repeat schedule functionality #676

UrekD:
Implement functionality discussed in #676:

- Option to define when task should end
- Implemented new logic to handle the ending (arg passing, json writing, schedule handling)
- Implemented new UI elements to select the end type

### DIFF
--- a/src/scheduling/schedule_config.py
+++ b/src/scheduling/schedule_config.py
@@ -42,9 +42,7 @@ def _read_end_arg_int(incoming_schedule_config):
 
 def _read_end_args(incoming_schedule_config):
     end_option = incoming_schedule_config.get('end_option')
-    if end_option is None:
-        raise InvalidScheduleException('end_option is required for repeatable schedule')
-    elif end_option == 'end_datetime':
+    if end_option == 'end_datetime':
         end_arg = _read_datetime(incoming_schedule_config, 'end_arg')
         return end_option,end_arg
     elif end_option == 'max_executions':
@@ -116,8 +114,6 @@ class ScheduleConfig:
         elif self.end_option == 'max_executions':
             result['end_option'] = self.end_option
             result['end_arg'] = self.end_arg
-        elif self.end_option == 'never':
-            result['end_option'] = 'never'
 
         if self.repeatable:
             result['executions_count'] = self.executions_count


bugy:
Hi @UrekD sorry for a late response. Unfortunately I'm a bit busy this week, so don't have time to review. Will try to check the pull request within the next several days.
And thanks a lot for the pull request!

src/model/external_model.py
```diff
@@ -129,6 +129,8 @@ def parse_external_schedule(external_schedule):
     return {
         'repeatable': external_schedule.get('repeatable'),
         'start_datetime': external_schedule.get('startDatetime'),
+        'endOption': external_schedule.get('endOption'),
+        'endArg': external_schedule.get('endArg'),
```
bugy, line 133: I would suggest using explicit names, e.g. `end_datetime` and `max_executions`, to make the logic more readable (in the files below).
Also, please use snake_case for naming in python

src/scheduling/schedule_config.py
```diff
@@ -52,10 +49,20 @@ def read_weekdays(incoming_schedule_config):
 
 def read_schedule_config(incoming_schedule_config):
     repeatable = read_repeatable_flag(incoming_schedule_config)
-    start_datetime = _read_start_datetime(incoming_schedule_config)
+    start_datetime = _read_datetime(incoming_schedule_config, 'start_datetime')
```
bugy, line 52: Please add validation for start_datetime is not None. It was a part of `_read_datetime` before, but you removed the validation

src/scheduling/schedule_config.py
```diff
@@ -52,10 +49,20 @@ def read_weekdays(incoming_schedule_config):
...
 
     prepared_schedule_config = ScheduleConfig(repeatable, start_datetime)
     if repeatable:
+
+        endOption = incoming_schedule_config.get('endOption')
+        prepared_schedule_config.endOption = endOption
```
bugy, line 58: Please use `snake_case`

src/scheduling/schedule_service.py
```diff
@@ -83,6 +83,15 @@ def create_job(self, script_name, parameter_values, incoming_schedule_config, us
         if not schedule_config.repeatable and date_utils.is_past(schedule_config.start_datetime):
             raise InvalidScheduleException('Start date should be in the future')
 
+        if schedule_config.endOption == 'on':
+            if schedule_config.start_datetime > schedule_config.endArg:
+                raise InvalidScheduleException('End date should be after start date')
+            if date_utils.is_past(schedule_config.endArg):
+                raise InvalidScheduleException('End date should be in the future')
```
bugy, line 90: I think this check is redundant. We already check, that start time is in the future

web-src/src/main-app/components/schedule/SchedulePanel.vue
```diff
@@ -5,62 +5,79 @@
       <div class="schedule-type-panel">
         <p class="schedule-type-field">
           <label>
-            <input :checked="oneTimeSchedule" class="with-gap" name="schedule-type"
-                   type="radio"
-                   @click="oneTimeSchedule = true"/>
+            <input :checked="oneTimeSchedule" class="with-gap" name="schedule-type" type="radio" @click="oneTimeSchedule = true" />
```
bugy, line 8: Could you revert changes in the existing lines please, which you didn't modify here? I really prefer my previous formatting (with new lines), rather than long lines, which are much harder to read.

src/scheduling/schedule_service.py
```diff
@@ -143,6 +160,14 @@ def cleanup():
                     self._execution_service.cleanup_execution(execution_id, user)
 
                 self._execution_service.add_finish_listener(cleanup, execution_id)
+
+            if job.schedule.endOption == 'after':
+                job.schedule.endArg -= 1
```
bugy, line 165: What do you think about not overriding the config file on every execution, but rather count how many executions could fit between now and start_datetime?
e.g. if `start_datetime=19:30`, `max_count=5`, `now=22:12`. Then until now there were 6 executions (19:30, 20:00, 20:30, 21:00, 21:30, 22:00) and there is no need to execute anymore? For this, we could reuse scheduling.schedule_config.ScheduleConfig.get_next_time

bugy (changes requested):

src/scheduling/schedule_service.py
UrekD: Sorry, I'm not really understanding what you mean here as start is in the past? I think executions shouldn't be tied to exact next execute time as this makes this option obsolete in my opinion as this can be achieve with end time. This option would be more of  a guarantee execution of the job x times so even downtime wouldn't effect the job.

src/scheduling/schedule_config.py
UrekD: My bad I'll re add, tho I belive currently it's not possible to even pass None as if you leave the fields empty it grabs the default values from page load as well date is not required.

image

src/model/external_model.py
UrekD: I believe it would be more logical to have it set up this way because you can use the "end_option" to determine which case should be executed (in if statements and similar situations). For example, if we wanted to run something for a job that doesn't have an end date, we would need to check that both the "end_datetime" and "max_executions" are set to None. It's not a big problem when there are only two options, but if another option is added in the future, it would require more work and fields in ScheduleConfig class.

web-src/src/main-app/components/schedule/SchedulePanel.vue
UrekD: Changes should be reverted.

src/scheduling/schedule_service.py
UrekD: Good point, removed.

src/scheduling/schedule_config.py
UrekD: Adjusted in all files.

bugy (changes requested):

web-src/src/main-app/components/schedule/SchedulePanel.vue
bugy: Thanks!

src/scheduling/schedule_config.py
bugy: thanks!

src/scheduling/schedule_service.py
bugy: Alright, makes sense. Another proposal: may be instead of overriding the maxCount, we could save one more parameter: executionsCount (i.e. how many times the script was executed). So that we will keep the original maxCount value as a source of truth

src/model/external_model.py
bugy: :+1: ok, agreed :)

src/scheduling/schedule_config.py
```diff
@@ -7,12 +7,11 @@
...
-def _read_start_datetime(incoming_schedule_config):
-    start_datetime = model_helper.read_datetime_from_config('start_datetime', incoming_schedule_config)
-    if start_datetime is None:
-        raise InvalidScheduleException('start_datetime is required')
-    return start_datetime
-
+def _read_datetime(incoming_schedule_config, key):
+    datetime_value = model_helper.read_datetime_from_config(key, incoming_schedule_config)
+    if datetime_value is None:
+        raise InvalidScheduleException('%1 is required', key)
```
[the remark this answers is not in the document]
UrekD: Indeed should have been f'..., fixed!

src/scheduling/schedule_service.py
```diff
@@ -113,8 +120,16 @@ def schedule_job(self, job: SchedulingJob, job_path):
...
             return
-
+        
+        if schedule.end_option == 'after' and schedule.end_arg <= 0:
+            return                
+        
         next_datetime = schedule.get_next_time()
+
+        if schedule.end_option == 'on':
+            if date_utils.is_past(schedule.end_arg) or next_datetime > schedule.end_arg:
```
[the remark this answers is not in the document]
UrekD: Yep checked and removed.

src/scheduling/schedule_config.py
```diff
@@ -52,10 +51,20 @@ def read_weekdays(incoming_schedule_config):
...
 
     prepared_schedule_config = ScheduleConfig(repeatable, start_datetime)
     if repeatable:
+
+        end_option = incoming_schedule_config.get('end_option')
+        prepared_schedule_config.end_option = end_option
+        if end_option == 'on':
+            prepared_schedule_config.end_arg = _read_datetime(incoming_schedule_config, 'end_arg')
+        elif end_option == 'after':
+            prepared_schedule_config.end_arg = model_helper.read_int_from_config('end_arg', incoming_schedule_config)
```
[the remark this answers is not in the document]
UrekD: Implemented

web-src/src/main-app/components/schedule/SchedulePanel.vue
```diff
@@ -93,12 +127,19 @@ export default {
...
     return {
       oneTimeSchedule: true,
       startDate: now,
       startTime: now.toTimeString().substr(0, 5),
+      endOption: 'never',
+      endDate: endDay,
+      endTime: endDay.toTimeString().substr(0, 5),
       id: null,
       repeatPeriod: 1,
+      executeCount: 1,
```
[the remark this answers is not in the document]
UrekD: Renamed

src/scheduling/schedule_config.py
```diff
@@ -52,10 +51,20 @@ def read_weekdays(incoming_schedule_config):
...
     repeatable = read_repeatable_flag(incoming_schedule_config)
-    start_datetime = _read_start_datetime(incoming_schedule_config)
+    start_datetime = _read_datetime(incoming_schedule_config, 'start_datetime')
 
     prepared_schedule_config = ScheduleConfig(repeatable, start_datetime)
     if repeatable:
+
+        end_option = incoming_schedule_config.get('end_option')
+        prepared_schedule_config.end_option = end_option
+        if end_option == 'on':
```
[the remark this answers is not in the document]
UrekD: Agreed, renamed em.

src/scheduling/schedule_service.py
UrekD: Great idea, how about including executionsCount  on all repeatable tasks to offer some statistic about how many times the job ran in general.

bugy: Sounds good to me

src/scheduling/schedule_config.py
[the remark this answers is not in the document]
UrekD: Seen the error in my ways, I have re done it to be more in line with the previous. I think when checking for execution count it doesn't make sense to error if it's not available but just to set it to 0.

src/scheduling/schedule_service.py
UrekD: Implemented.

src/scheduling/schedule_config.py
```diff
@@ -32,6 +31,37 @@ def _read_repeat_period(incoming_schedule_config):
     return period
 
 
+def _read_executions_count(incoming_schedule_config):
+    executions_count = model_helper.read_int_from_config('executions_count', incoming_schedule_config)
```
bugy, line 35: You could pass `default` parameter to `read_int_from_config`

src/scheduling/schedule_service.py
```diff
@@ -136,6 +151,14 @@ def cleanup():
                     self._execution_service.cleanup_execution(execution_id, user)
 
                 self._execution_service.add_finish_listener(cleanup, execution_id)
+
+            if job.schedule.repeatable:
+                job.schedule.executions_count += 1
+
+                file_utils.write_file(
```
bugy, line 158: Can we call `save_job` ?

src/scheduling/schedule_config.py
```diff
@@ -32,6 +31,37 @@ def _read_repeat_period(incoming_schedule_config):
...
+        raise InvalidScheduleException('end_arg is required for repeatable schedule')
+    elif end_arg <= 0:
+        raise InvalidScheduleException('end_arg should be > 0')
+    return end_arg
+
+
+def _read_end_args(incoming_schedule_config):
+    end_option = incoming_schedule_config.get('end_option')
+    if end_option is None:
+        raise InvalidScheduleException('end_option is required for repeatable schedule')
```
bugy, line 54: I think None `end_option` should be allowed:
- existing configurations don't have it
- we should make manual configuration (via json) more user-friendly, i.e. don't require unnecessary configurations

src/scheduling/schedule_config.py
```diff
@@ -80,6 +118,18 @@ def as_serializable_dict(self):
...
         }
 
+        if self.end_option == 'end_datetime':
+            result['end_option'] = self.end_option
+            result['end_arg'] = date_utils.to_iso_string(self.end_arg)
+        elif self.end_option == 'max_executions':
+            result['end_option'] = self.end_option
+            result['end_arg'] = self.end_arg
+        elif self.end_option == 'never':
+            result['end_option'] = 'never'
```
bugy, line 128: Should we just skip it? As in a comment above, I would suggest to completely omit `end_...` configuration if it's not needed.

bugy (changes requested):
Hi @UrekD I think we are almost there :) I have only one big concern, about making this new feature optional

UrekD:
> Hi @UrekD I think we are almost there :) I have only one big concern, about making this new feature optional

Hey @bugy, how so? I'm guessing maybe from an administrative standpoint? As the functionality shouldn't in no way hinder the functionality as it was before, if you don't select anything else. I see the point about not requiring a value for never ending schedules that was oversight on my end, which actually thought about not implementing in a current way but not for that reason :D I'll address the new stuff tomorrow, thanks!

bugy (approved):
Sorry for delay and lots of previous comments
Looks good to me now, thanks a lot for the contribution! :+1: :tada:

UrekD:
> Sorry for delay and lots of previous comments Looks good to me now, thanks a lot for the contribution! 👍 🎉

No worries, thank you!

vnghia:
Can we avoid overwriting the configuration file ? Because I am using an Infra-as-code approach and everything is generated by some kind of code and set to read-only afterward. So this failed on my server.

UrekD:
> Can we avoid overwriting the configuration file ? Because I am using an Infra-as-code approach and everything is generated by some kind of code and set to read-only afterward. So this failed on my server.
> 
> https://github.com/bugy/script-server/blob/7ea018023532d42b101fef2ebf7b49ded2ec6b1a/src/scheduling/schedule_service.py#L155-L158
> 
> Maybe add a `is_forever` attribute to the `schedule_config` and we won't touch the file again if it has `is_forever`. I can send a PR for that.

Yeah that sounds good maybe better calling it `allow_write`.  Personally I'd use a modified image for deployment, as with the new var implementation you will also need to handle a schedule that is set to count for it to not fire forever if it can't keep track of the count, not sure if it's worth the hassle.

PR #700, opened 2023-09-29
Make attachments optional for email alerts

boblickj:
One downstream system we use does not play nice with emails containing attachments. This PR adds the ability to disable attachments when sending email alerts. The default action for users who specify nothing does not change.

### DIFF
--- a/src/communications/destination_email.py
+++ b/src/communications/destination_email.py
@@ -56,6 +56,7 @@ class EmailCommunicator:
         self.auth_enabled = read_bool_from_config('auth_enabled', params_dict)
         self.login = params_dict.get('login')
         self.tls = read_bool_from_config('tls', params_dict)
+        self.include_files = read_bool_from_config('include_files', params_dict, default=True)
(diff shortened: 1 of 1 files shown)


src/communications/destination_email.py
```diff
@@ -56,6 +56,7 @@ def __init__(self, params_dict):
         self.auth_enabled = read_bool_from_config('auth_enabled', params_dict)
         self.login = params_dict.get('login')
         self.tls = read_bool_from_config('tls', params_dict)
+        self.include_files = read_bool_from_config('include_files', params_dict, default=True)
```
bugy, line 59: can we use name "attach_files" instead, please?

bugy (changes requested):

src/communications/destination_email.py
boblickj: sure, i agree that fits better

bugy (approved):
Thank you!

PR #730, opened 2024-01-19
fix: XSS attack via next login parameter.

yog27ray:
Vulnerable Parameter:
?next=

Payload Used:
javascript:prompt(document.domain);//

Steps to Reproduce the Bug:
1. Visit http://server.url/login.html?next=javascript:prompt(document.domain);//
2. Click on the "Sign in with Google" button.
3. Login into your Account.
4. XSS will be triggered.


bugy (approved):
thank you!

Issue #741, opened 2024-03-25
feature(auth): Azure AD OAuth


0lionelzhang0:
Hello!

Thank you bugy for this amazing tool. I was wondering if there are any plans to add auth support for Azure AD OAuth?

If not, would you mind providing some pointers on how one could implement that?

Thanks!

bugy:
Hi, unfortunately, it's not supported. But if it's more or less the standard OAuth implementation, it should be quite straightforward to implement.
Here is an example of Google OAuth: https://github.com/bugy/script-server/blob/master/src/auth/auth_google_oauth.py
It's pretty small, since the main implementation is located in the parent OAuth class: https://github.com/bugy/script-server/blob/master/src/auth/auth_abstract_oauth.py

Besides this class, the following things should be updated:
- https://github.com/bugy/script-server/blob/6015e14bfc286f0ad48ad61671706112672a6fee/src/model/server_conf.py#L222 (add an initialization for a new type)
- https://github.com/bugy/script-server/blob/master/web-src/src/login/login.js (UI component, responsible for login flow)

0lionelzhang0:
Thank you, bugy. I will look into it!

PR #745, opened 2024-04-01
feat: add support for azure ad oauth

0lionelzhang0:
With regards to #741
Adds support for Azure AD OAuth

- requires the following config:
  - auth.type (azure_ad_oauth)
  - auth.auth_url (https://login.microsoftonline.com/<TENANT_ID>/oauth2/v2.0/authorize)
  - auth.token_url (https://login.microsoftonline.com/<TENANT_ID>/oauth2/v2.0/token)
  - auth.client_id
  - auth.secret


bugy:
I'm so sorry for missing it. Will review it now

src/auth/auth_azure_ad_oauth.py
```diff
@@ -0,0 +1,34 @@
...
+        super().__init__(
+            self.auth_url,
+            self.token_url,
+            'openid email profile',
+            params_dict,
+        )
+
+    async def fetch_user_info(self, access_token) -> _OauthUserInfo:
+        headers = {'Authorization': f'Bearer {access_token}'}
+        user_response = await self.http_client.fetch('https://graph.microsoft.com/v1.0/me', headers=headers)
```
bugy, line 26: did you try using the same approach as in GitlabOAuthAuthenticator? I.e. extend `OAuth2Mixin` and then make a request by
```
user = await self.oauth2_request(
            'https://graph.microsoft.com/v1.0/me',
            access_token=access_token)
```

bugy, line 26: Is the host always `'https://graph.microsoft.com/v1.0/me'` or it can be in a private namespace or on premise?

src/auth/auth_azure_ad_oauth.py
```diff
@@ -0,0 +1,34 @@
...
+from auth.auth_abstract_oauth import AbstractOauthAuthenticator, _OauthUserInfo
+from model import model_helper
+
+LOGGER = logging.getLogger('script_server.AzureADOauthAuthenticator')
+
+
+class AzureAdOAuthAuthenticator(AbstractOauthAuthenticator):
+    def __init__(self, params_dict):
+        params_dict['group_support'] = False
+        self.auth_url = model_helper.read_obligatory(params_dict, 'auth_url', ' for OAuth')
```
bugy, line 14: May be we should ask only for `tenant_id` from the config? Since the URLs always have the same structure

0lionelzhang0: I believe these URLs can be copy-pasted directly from the Azure Portal when a new app is registered. So I guess to semi-future proof it, we can leave the full URL in the config in case Microsoft changes the URL structure.

For reference, I was mostly going by the [Grafana Azure AD setup instructions](https://grafana.com/docs/grafana/latest/setup-grafana/configure-security/configure-authentication/azuread/#create-the-microsoft-entra-id-application)

src/auth/auth_azure_ad_oauth.py
0lionelzhang0: Yes, my understanding is the URL is always this for Azure AD until v2.0 comes out. [When v2.0 comes out, the v1.0 endpoint will remain supported for 2 years following the announcement](https://learn.microsoft.com/en-us/graph/versioning-and-support).

Azure AD is strictly cloud-based, a private/on-prem setup would be using [Active Directory Domain Services](https://learn.microsoft.com/en-us/windows-server/identity/ad-ds/get-started/virtual-dc/active-directory-domain-services-overview) instead.

0lionelzhang0: Thanks, I will try this next week!

0lionelzhang0:
Authentication.md wiki snippet:

```
## Azure AD OAuth

Script server can authenticate users via Azure AD OAuth. `auth_url`, `token_url`, `client_id`, and `secret` properties should be acquired from Azure and configured in the Script server.

### Instructions
- Log in to [Azure Portal](https://portal.azure.com).
- Click **Microsoft Entra ID** in the side menu. If you have access to more than one tenant, select your account in the upper right and set your session to the Entra ID tenant you wish to use.
- Under **Manage** in the side menu, click **App Registrations** > **New Registration**.
- Enter a descriptive name.
- Under **Redirect URI**, select the app type **Web**.
- Add the following redirect URLs:
  - `<script-server-url>/login.html`
  - `<script-server-url>`
- Click **Register**. The app’s Overview page opens.
- Note the **Application ID**. This is the OAuth client ID (`client_id`).
- Click **Endpoints** from the top menu.
  - Note the OAuth 2.0 authorization endpoint (v2) URL (`auth_url`).
  - Note the OAuth 2.0 token endpoint (v2) URL (`token_url`).
- Click **Certificates & secrets** in the side menu.
- Add a new entry under **Client secrets** with the following configuration:
  - **Description**: Grafana OAuth
  - **Expires**: Select an expiration period
- Click **Add** then copy the key value. This is the OAuth client secret (`secret`).

### Options
* `type` (required) - should be `azure_ad_oauth`
* `auth_url` (required) - e.g. `https://login.microsoftonline.com/<TENANT_ID>/oauth2/v2.0/authorize`
* `token_url` (required) - e.g. `https://login.microsoftonline.com/<TENANT_ID>/oauth2/v2.0/token`
* `client_id` (required) - OAuth client ID
* `secret` (required) - OAuth client secret
```

src/auth/auth_azure_ad_oauth.py
bugy: thanks!

src/auth/auth_azure_ad_oauth.py
bugy: ok, fair, thanks!

src/auth/auth_azure_ad_oauth.py
0lionelzhang0: Unfortunately, I wasn't able to get it to work with OAuth2Mixin. I would get unauthorized error 401. It might be that the Microsoft endpoint doesn't support the fetch methods of either embedding access token in URL or access token as POST request body.
https://learn.microsoft.com/en-us/graph/auth-v2-user?tabs=http#request

^ Their example in this link only shows using the header method.

bugy: Thanks for trying!

tvalenta:
@bugy is there a dev release of this anywhere that can be downloaded?

bugy:
https://github.com/bugy/script-server/releases/tag/dev

tvalenta:
The dev release has a date of December 2023. Am I misreading it?

bugy:
@tvalenta oh, the build was not running for a long time already :(
I sent a message to travis team to provide more build credits, so the builds can run again.

tvalenta:
> @tvalenta oh, the build was not running for a long time already :( I sent a message to travis team to provide more build credits, so the builds can run again.

I'd like to give this build a try -- any chance cicd would now build the dev release?

bugy:
@tvalenta the build should be ready now :) sorry for delay

Issue #788, opened 2025-04-22
Log files add a parameter to each script


AlanMerry:
Currently there appears to be no way to control the generation of log files.

I have a job running every 5 mins.

I do not want to see the log files.

Can we have a new parameter which controls the calling of the logging script ?

MNeill73:
The ability to disable the logging opens up the possibility of bad actors
installing scripts that you do want logged that don't.

The logfile names are based on the script invocation. Just run a crown job
every (5/10/15) minutes to purge those logs.

AlanMerry:
The default for scripts would be to create logs.

The question is to have a parameter per script, new unknown scripts would still create logs.

bugy:
Hi @AlanMerry I won't be able to do it, but you can submit a PR for it, if you have time.

bugy:
actually I tried to usecopilot to solve this, let's see the results :)

PR #789, opened 2025-04-23
#788: added possibilit to disable logging per script

bugy:
#788: added possibilit to disable logging per script


bugy:
Test results: https://script-server-tests.s3-us-west-2.amazonaws.com/master/780/index.html

PR #812, opened 2026-02-19
Add secure and httponly options for cookies

cpadlab:
I have added the option to configure Secure and HttpOnly flags for cookies, as this prevents potential cookie theft or unauthorized access via scripts.

I have been testing this implementation locally on a server, and it works perfectly. To use it, the user only has to update the security section in the configuration file.

---

Example change in `conf/conf.json`:

From this:

```
"security": {
    "xsrf_protection": "token"
}
```

To this:

```
"security": {
    "xsrf_protection": "header",
    "cookie_secure": true,
    "cookie_httponly": true
}
```

### DIFF
--- a/src/auth/identification.py
+++ b/src/auth/identification.py
@@ -117,8 +117,7 @@ class IpBasedIdentification(Identification):
         expiry_time = date_utils.get_current_millis() + days_to_ms(self.EXPIRES_DAYS)
         new_token = client_id + '&' + str(expiry_time)
         server_config = request_handler.application.server_config
-        request_handler.set_secure_cookie(
-            self.COOKIE_KEY, new_token, expires_days=self.EXPIRES_DAYS, secure=server_config.cookie_secure, httponly=server_config.cookie_httponly)
+        request_handler.set_secure_cookie(self.COOKIE_KEY, new_token, expires_days=self.EXPIRES_DAYS, secure=server_config.cookie_secure, httponly=server_config.cookie_httponly)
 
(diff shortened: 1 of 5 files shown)


bugy:
Hi @cpadlab thanks!
Do you see any reason to make it configurable? 
Can we just set these parameters for everyone?

cpadlab:
> Hi @cpadlab thanks! Do you see any reason to make it configurable? Can we just set these parameters for everyone?

Hi, I wanted to make it configurable to avoid issues with users deploying it on HTTP servers

bugy:
I see. I guess we can set "httponly" to true always
For `secure` attribute you are right. But I would still make the default option as true and users would need to opt in for unsecure cookies.

What do you think?

cpadlab:
Hi @bugy , You're absolutely right, I've just pushed the new commits

bugy (approved):

bugy:
Sorry for the delay, thanks a lot!

etiennecaldichouryatnapta:
Hi team, 
Since I updated script-server I get sometimes this error. Looks like this MR could be the root cause?

image

image

bugy:
Hi @etiennecaldichouryatnapta thanks for reporting, i committed a fix, can you try again? 

PS i noticed that Travis is not running due to missing credits. I contacted them to get more credits and run builds again

etiennecaldichouryatnapta:
Hi @bugy thank you very much for your answer and your fix, we'll try it ASAP and come back to you

PR #821, opened 2026-06-24
feat: make `web` folder configurable

duesee:
This PR adds a parameter to make the `web` folder configurable. This has a few benefits, one is that we can easily experiment with different web roots. If I didn't overlook something, not using the parameter (which defaults to "web") should keep the current behavior.

Note: This PR is mainly based on https://github.com/tfc/script-server.

### DIFF
--- a/src/utils/tool_utils.py
+++ b/src/utils/tool_utils.py
@@ -3,9 +3,7 @@ import os
 from utils import file_utils
 
 
-def validate_web_build_exists(project_path):
-    web_folder = os.path.join(project_path, 'web')
-
+def validate_web_build_exists(web_folder):
     how_to_fix_build_message = \
         'How to fix: ' \
         '\n - PROD: please use stable releases from https://github.com/bugy/script-server/releases/latest' \
--- a/src/main.py
+++ b/src/main.py
@@ -30,10 +30,12 @@ parser.add_argument('-d', '--config-dir', default='conf')
 parser.add_argument('-f', '--config-file', default='conf.json')
 parser.add_argument('-l', '--log-folder', default='logs')
 parser.add_argument('-t', '--tmp-folder', default='temp')
+parser.add_argument('-w', '--web-folder', default='web')
 args = vars(parser.parse_args())
 
 TEMP_FOLDER = args['tmp_folder']
 LOG_FOLDER = args['log_folder']
+WEB_FOLDER = args['web_folder']
 
 CONFIG_FOLDER = args['config_dir']
 if os.path.isabs(args['config_file']):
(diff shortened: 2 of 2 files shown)


src/utils/tool_utils.py
```diff
@@ -3,9 +3,7 @@
 from utils import file_utils
 
 
-def validate_web_build_exists(project_path):
-    web_folder = os.path.join(project_path, 'web')
-
+def validate_web_build_exists(web_folder):
```
duesee, line 6: On a closer look, the difference now is that we don't print the whole path. We could use `os.path.realpath(web_folder)`, what do you think?

bugy:
Hey @duesee thanks for your contribution. Can you explain your use case please?
Why is it really needed?

duesee:
Sorry, yes, sure! From my understanding, the frontend (`web-src`) is a totally fine standalone package. In theory, it could be bundled and hosted (statically) anywhere, e.g., via nginx, a CDN, etc.

For us, it made sense to package it up separately with Nix so that we can refer from any other package to it, it doesn't need to be rebuild when the backend changes, we can quickly patch it if needed, etc.

In Nix, however, packages are all put under `/nix/store/<hash>-<package>`. So we needed a way to tell tornado where to find it (because when build like this, it won't live next to the Python application).

The other benefit is that it makes testing a little easier since we can quickly change the web source via the parameter.

Having said that... if you feel this is introducing complexity that you don't like, we can work around it! But in my opinion the increased flexibility outweights the complexity.

bugy:
From my perspective, BE and FE are tightly coupled. I cannot imagine a situation when they would be used as standalone modules. 

May be you can explain Nix structure a bit more. How these 2 separate packages would live side by side. What would be the release cycle and in general how released for Nix would look like.

>  if you feel this is introducing complexity that you don't like

I don't like having features, which I don't understand :) It's hard to maintain something what you don't know. That's why I'm asking additional questions to learn more about it.

duesee:
I understand :-) Hm... maybe it makes sense then for me to provide a PR with the required Nix files as a basis for discussion? I could submit a (draft) PR so we have something to look on.

(I probably won't use my computer the next days, but will surely submit something next week :-))

bugy:
sure, thanks!

duesee:
I still think having the possibility to change the web root is worth it. But I'll let you decide what direction to take :-) Feel free to close this PR (or, fingers-crossed, tell me how to push it over the finish line :-))

bugy (approved):

bugy:
let's merge it :)

duesee:
Hello! Since we agreed to merge... should we do? :-) Maybe CI needs a restart?

bugy:
Hi, sorry, of course
No idea why i didn't do it
May be got distracted and forgot

PR #823, opened 2026-07-06
build: Add Nix/NixOS support

duesee:
Hello, and sorry for the delay!

Here comes the first version (where I avoided [depending on a configurable `web` folder](https://github.com/bugy/script-server/pull/821).) I've added a few comments inline.

# Why do we want it?

## NixOS integration

With the files in place, NixOS users can (basically) add ...

```nix
services.script-server.enable = true;
```

... to their NixOS configuration and get a fully configured (systemd) script-server.

## Linux VM demo

With the files in place, Nix users can start a live qemu VM demo with script-server preconfigured via ...

```nix
nix run .#demo
```

## Linux image creation

With the files in place, Nix users can create {qemu,lxc,oci,...} images with script-server preinstalled and preconfigured by running ...

```nix
nix run .#nixosConfigurations.demo.config.system.build.images.<variant>
```

Especially LXC containers could make a lot of sense since they can be run natively on Proxmox.
<details>

<summary>Full list of supported images</summary>

```text
nixosConfigurations.demo.config.system.build.images.amazon              nixosConfigurations.demo.config.system.build.images.kubevirt            nixosConfigurations.demo.config.system.build.images.qemu
nixosConfigurations.demo.config.system.build.images.azure               nixosConfigurations.demo.config.system.build.images.linode              nixosConfigurations.demo.config.system.build.images.qemu-efi
nixosConfigurations.demo.config.system.build.images.cloudstack          nixosConfigurations.demo.config.system.build.images.lxc                 nixosConfigurations.demo.config.system.build.images.raw
nixosConfigurations.demo.config.system.build.images.digital-ocean       nixosConfigurations.demo.config.system.build.images.lxc-metadata        nixosConfigurations.demo.config.system.build.images.raw-efi
nixosConfigurations.demo.config.system.build.images.google-compute      nixosConfigurations.demo.config.system.build.images.oci                 nixosConfigurations.demo.config.system.build.images.sd-card
nixosConfigurations.demo.config.system.build.images.hyperv              nixosConfigurations.demo.config.system.build.images.openstack           nixosConfigurations.demo.config.system.build.images.vagrant-virtualbox
nixosConfigurations.demo.config.system.build.images.iso                 nixosConfigurations.demo.config.system.build.images.openstack-zfs       nixosConfigurations.demo.config.system.build.images.virtualbox
nixosConfigurations.demo.config.system.build.images.iso-installer       nixosConfigurations.demo.config.system.build.images.proxmox             nixosConfigurations.demo.config.system.build.images.vmware
nixosConfigurations.demo.config.system.build.images.kexec               nixosConfigurations.demo.config.system.build.images.proxmox-lxc
```
</details>

## NixOS test integration

With the files in place (and a little more effort), we could provide end-to-end integration tests where we automatically spawn a script-server via qemu and run all kinds of integration tests.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -80,3 +80,7 @@ web-src/geckodriver.log
 venv/
 /venv2/
 e2e_venv/
+
+# Nix / NixOS
+/result
+/*.qcow2
--- a/nix/module.nix
+++ b/nix/module.nix
@@ -1,0 +1,75 @@
+{
+  config,
+  pkgs,
+  lib,
+  ...
+}:
+let
+  cfg = config.services.script-server;
+  json = pkgs.formats.json { };
+in
+{
+  options.services.script-server = {
+    enable = lib.mkEnableOption "Whether to enable script-server.";
+
+    package = lib.mkOption {
+      type = lib.types.package;
+      default = pkgs.script-server;
+      description = "Script-server package to use.";
+    };
+
+    package-dist = lib.mkOption {
+      type = lib.types.package;
+      default = pkgs.script-server-dist;
+      description = "Script-server-dist package to use.";
+    };
+
+    settings = lib.mkOption {
+      type = json.type;
+      default = { };
+      description = "Server configuration, see <https://github.com/bugy/script-server/wiki/Server-configuration>.";
+    };
+
+    configuration = lib.mkOption {
+      type = lib.types.str;
+      default = "/var/lib/script-server/conf";
+      description = "Script configuration, see <https://github.com/bugy/script-server/wiki/Script-config>.";
+    };
+  };
+
+  config = lib.mkIf cfg.enable {
+    environment.systemPackages = [ cfg.package ];
+
+    environment.etc."script-server/conf.json".source =
+      json.generate "script-server-config" cfg.settings;
+
+    # See <https://github.com/bugy/script-server/wiki/Running-as-a-linux-service#systemd>
+    systemd.services.script-server = {
+      enable = true;
+
+      description = "Script Server";
+      after = [ "network.target" ];
+      wantedBy = [ "multi-user.target" ];
+
+      serviceConfig = {
+        Restart = "always";
+        RestartSec = "1s";
+        ExecStart = ''
+          ${cfg.package}/bin/script-server \
+            --config-file /etc/script-server/conf.json \
+            --config-dir ${cfg.configuration} \
+            --log-folder /var/log/script-server \
+            --tmp-folder /tmp
+        '';
+
+        # `main.py` expects `web` to be in working directory.
+        WorkingDirectory = "${cfg.package-dist}";
+        StateDirectory = "script-server";
+        LogsDirectory = "script-server";
+
+        DynamicUser = true;
+        PrivateTmp = true;
+      };
+    };
+  };
+}
--- a/nix/package-dist.nix
+++ b/nix/package-dist.nix
@@ -1,0 +1,23 @@
+{ lib, buildNpmPackage }:
+let
+  version = "1.18.0";
+in
+buildNpmPackage {
+  pname = "script-server-dist";
+  inherit version;
+
+  src = ../web-src;
+
+  npmDepsHash = "sha256-aBZDh2NcDU4OpsdTS8JqwGMa8WeIoyW9I6bUwTXfllU=";
+
+  makeCacheWritable = true;
+
+  # See <https://stackoverflow.com/questions/75959563/node-js-err-ossl-evp-unsupported-error-when-running-npm-run-start>
+  NODE_OPTIONS = "--openssl-legacy-provider";
+
+  installPhase = ''
+    mkdir "$out"
+    echo '${version}' > "$out/version.txt"
+    cp -r ../web "$out"
+  '';
+}
--- a/nix/package.nix
+++ b/nix/package.nix
@@ -1,0 +1,30 @@
+let
+  toml = builtins.fromTOML (builtins.readFile ./../pyproject.toml);
+in
+{
+  lib,
+  buildPythonPackage,
+  tornado,
+  setuptools,
+}:
+buildPythonPackage {
+  pname = toml.project.name;
+  version = toml.project.version;
+  pyproject = true;
+
+  src = lib.fileset.toSource {
+    root = ./..;
+    fileset = lib.fileset.unions [
+      ./../pyproject.toml
+      ./../src
+    ];
+  };
+
+  build-system = [
+    setuptools
+  ];
+
+  dependencies = [
+    tornado
+  ];
+}
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,61 @@
+{
+  "nodes": {
+    "flake-parts": {
+      "inputs": {
+        "nixpkgs-lib": "nixpkgs-lib"
+      },
+      "locked": {
+        "lastModified": 1782949081,
+        "narHash": "sha256-vp6Y/Grm98ESt6ceOkWiHWyZRDV3J1RID4w+6NWK9yA=",
+        "owner": "hercules-ci",
+        "repo": "flake-parts",
+        "rev": "17c9d6cdfc60c64f4ee8d306f9bc0b4ccb51481e",
+        "type": "github"
+      },
+      "original": {
+        "owner": "hercules-ci",
+        "repo": "flake-parts",
+        "type": "github"
+      }
+    },
+    "nixpkgs": {
+      "locked": {
+        "lastModified": 1780145889,
+        "narHash": "sha256-md0zn0RnwNvPyASas1yG5YUuwQ4ALA6ucL50l0DvqCo=",
+        "owner": "NixOS",
+        "repo": "nixpkgs",
+        "rev": "8c50a710ddca43d7a530fb805ad55bde8d0141c5",
+        "type": "github"
+      },
+      "original": {
+        "owner": "NixOS",
+        "ref": "26.05",
+        "repo": "nixpkgs",
+        "type": "github"
+      }
+    },
+    "nixpkgs-lib": {
+      "locked": {
+        "lastModified": 1782614948,
+        "narHash": "sha256-ePjCwr1sNm9NYUqywL7QfK3JnlS015msC+eBu2zKlp8=",
+        "owner": "nix-community",
+        "repo": "nixpkgs.lib",
+        "rev": "db3f255737b94216eb71cce308e2912cf6bc2d7c",
+        "type": "github"
+      },
+      "original": {
+        "owner": "nix-community",
+        "repo": "nixpkgs.lib",
+        "type": "github"
+      }
+    },
+    "root": {
+      "inputs": {
+        "flake-parts": "flake-parts",
+        "nixpkgs": "nixpkgs"
+      }
+    }
+  },
+  "root": "root",
+  "version": 7
+}
--- a/flake.nix
+++ b/flake.nix
@@ -1,0 +1,46 @@
+{
+  description = "Script-server";
+
+  inputs = {
+    nixpkgs.url = "github:NixOS/nixpkgs/26.05";
+
+    flake-parts.url = "github:hercules-ci/flake-parts";
+  };
+
+  outputs =
+    inputs@{
+      self,
+      nixpkgs,
+      flake-parts,
+      ...
+    }:
+    flake-parts.lib.mkFlake { inherit inputs; } {
+      flake = {
+        overlays.default = final: prev: {
+          script-server = final.python3Packages.callPackage ./nix/package.nix { };
+
+          script-server-dist = final.callPackage ./nix/package-dist.nix { };
+        };
+
+        nixosModules.default = import ./nix/module.nix;
+      };
+
+      systems = [
+        "x86_64-linux"
+        "aarch64-linux"
+        "aarch64-darwin"
+      ];
+
+      perSystem =
+        { pkgs, ... }:
+        {
+          packages = {
+            script-server = pkgs.python3Packages.callPackage ./nix/package.nix { };
+
+            script-server-dist = pkgs.callPackage ./nix/package-dist.nix { };
+          };
+
+          formatter = pkgs.nixfmt-tree;
+        };
+    };
+}


nix/module.nix
duesee, line 1: Defines how the script-server would be configured in NixOS.

With `enable = true;` we "install" the script-server. `settings` is a Nix "attribute set" / dictionary that is placed unchanged in JSON format to `/etc/script-server/conf.json`. `configuration` is the path to the script configuration.

Example:

```nix
services.script-server = {
  enable = true;

  # Optional
  settings = { }; # JSON config in Nix syntax
  # Optional
  configuration = "..."; # Path to `conf` / `database` folder; Default: "/var/lib/script-server/conf"
};
```

nix/module.nix
```diff
@@ -0,0 +1,75 @@
...
+      type = lib.types.str;
+      default = "/var/lib/script-server/conf";
+      description = "Script configuration, see <https://github.com/bugy/script-server/wiki/Script-config>.";
+    };
+  };
+
+  config = lib.mkIf cfg.enable {
+    environment.systemPackages = [ cfg.package ];
+
+    environment.etc."script-server/conf.json".source =
```
duesee, line 43: "Settings" could be put under `/etc/script-server/conf.json` when using NixOS.

nix/module.nix
```diff
@@ -0,0 +1,75 @@
...
+        RestartSec = "1s";
+        ExecStart = ''
+          ${cfg.package}/bin/script-server \
+            --config-file /etc/script-server/conf.json \
+            --config-dir ${cfg.configuration} \
+            --log-folder /var/log/script-server \
+            --tmp-folder /tmp
+        '';
+
+        # `main.py` expects `web` to be in working directory.
```
duesee, line 65: From my understanding: When `launcher.py` is used, the `web` folder must be co-located to it. We could bundle the NixOS package so that this is always the case and keep calling `launcher.py`. But... according to the [Filesystem Hierarchy Standard](https://refspecs.linuxfoundation.org/FHS_3.0/fhs/index.html), we probably should put `launcher.py` under `/usr/bin/launcher.py` and would then need to put the `web` folder there as well which feels off.

Instead, we call `main.py` and set the working directory to where the `web` package is located in the systemd unit. It feels hacky and I think we need to put `version.txt` there as well...

(I think this is not specific to NixOS. Packaging for e.g. Ubuntu would raise the same questions?)

nix/package.nix
duesee, line 1: This builds the backend.

nix/package-dist.nix
duesee, line 1: This builds/bundles the frontend.

.gitignore
```diff
@@ -80,3 +80,7 @@ web-src/geckodriver.log
 venv/
 /venv2/
 e2e_venv/
+
+# Nix / NixOS
```
duesee, line 84: `result` is for Nix build artifacts. There is also a qemu demo startable via `nix run .#demo`. Since the demo will instantiate a VM, we ignore the `.qcow2` hard drive, too.

bugy:
How is it supposed to be used/built?

Imagine that I'm a normal Nix user (_which would be a bad example, since I have no clue about nix_), who would like to start using script server.
What should I do?

duesee:
If you have Nix installed (see https://nixos.org/download/), `cd` into the script-server folder, and execute `nix run .#demo` (to start a qemu VM), `nix run .#nixosConfigurations.demo.config.system.build.images.lxc` (to e.g. build an lxc image), `nix build .#script-server-frontend` to bundle the `web` sources, etc.

If you are on NixOS system, you could (almost*) only add `services.script-server.enable = true;` to your global system configuration to install and preconfigure script-server on your machine.

You can look at, e.g., https://wiki.nixos.org/wiki/Nginx to get a feeling for NixOS. You could think about it being a "wrapper" over everything Linux/packages that allows you to fully declaratively and reproducibly define your running system. If you decide later to not wanting a software/service anymore, you delete the lines from your config and everything is cleanly rebuild w/o the software/service from scratch.

f you feel this is going in the right direction, I would love to add a little to the `README`.

*) We currently also need to point the system to this repo, but let's see how far we come :-)

nix/module.nix
bugy: But working dir is used not only for web, is it? All other folders from the release are expected to be there

nix/package-dist.nix
```diff
@@ -0,0 +1,19 @@
+{ lib, buildNpmPackage }:
+buildNpmPackage {
+  pname = "script-server-web";
+  version = "1.18.0";
+
+  src = ../web-src;
+
+  npmDepsHash = "sha256-aBZDh2NcDU4OpsdTS8JqwGMa8WeIoyW9I6bUwTXfllU=";
```
bugy, line 11: Can it be automatically calculated as a result of some build tool?

bugy: Actually it would be the same for "version"

bugy (changes requested):
In general looks fine to me (considering that I have no clue about nix ecosystem)
But I would highly prefer to put all these files into tools/nix folder
To avoid polluting the main folder
Also, I would rather put "demo" config (if possible) into wiki section of github

nix/package-dist.nix
duesee: The hash ensures that we always downloaded the exact same npm dependencies. We could adapt the hash in CI when updating the deps, though, to make our life easier. But I would suggest to tackle this later.

If you don't mind, I'll take care about the maintenance of the Nix support. Ideally, the files can eventually be moved into `nixpkgs` but I don't have experience here and probably want to ask someone before starting anything :-) Let's see first how well this works.

Since you seem fine with the direction, I'll try to finalize the PR soon(tm) and write a short NixOS Wiki entry.

duesee:
```nix
soon = "probably next week";
```

nix/module.nix
duesee: From what I know/see, it is `web` and `version.txt`. I made the `version.txt` work, too.

duesee:
Hey and thanks for the patience! I think this should be fine now (and would already help us very much.)

I would propose to put a minimal (but still complete) example into the docs. Nix/NixOS users will know how to integrate it into their system config and use it.

```nix
# flake.nix
{
  inputs = {
    nixpkgs.url = "github:NixOS/nixpkgs/26.05";
    script-server.url = "github:bugy/script-server";
  };

  outputs = { nixpkgs, script-server, ... }: {
    nixosConfigurations = {
      demo = nixpkgs.lib.nixosSystem {
        modules = [
          {
            imports = [
              script-server.nixosModules.default
            ];

            nixpkgs = {
              overlays = [ script-server.overlays.default ];
              hostPlatform = "x86_64-linux";
            };

            networking.firewall.allowedTCPPorts = [ 5000 ];

            services = {
              script-server = {
                enable = true;
                settings = {
                  title = "Service Center";
                };
                configuration = "/var/lib/script-server/conf";
              };
            };
          }
        ];
      };
    };
  };
}
```

My proposal would be to go with it as it is for a while and see how it works out. As soon as we get more experience and start to feel confident that everything works as we want, we could try to merge it into `nixpkgs`. This way, much of the code from above will go away and script-server could basically really be installed by doing `services.script-server.enable = true;` But: Future work :-)

nix/package-dist.nix
duesee, line 1: I renamed `script-server-web` to `script-server-dist` and it now auto-generates the `version.txt`. This way, the version is also showed correctly.

duesee:
Hey @bugy, is there anything left for me to do here? No hurry, just want to know if I can improve something in the meantime.

bugy (changes requested):
Thanks a lot for this, and no worries about the delay! The packaging itself is clean and idiomatic, and I like that you avoided the configurable-`web`-folder dependency from #821.

My main concern is the NixOS module: a few of script-server's runtime assumptions aren't met by the unit as written, so a fresh `services.script-server.enable = true;` won't come up. Details below, roughly in priority order.

## Blocking

### 1. Fresh install will crash-loop — `conf/logging.json` is never provisioned

`main()` opens `<config-dir>/logging.json` unconditionally (`src/main.py:69-71`), *before* `file_utils.prepare_folder(CONFIG_FOLDER)` runs. The module points `--config-dir` at `/var/lib/script-server/conf`, which starts out empty, and `conf/logging.json` isn't in either package's fileset (`package.nix` ships only `pyproject.toml` + `src`; `package-dist.nix` only `web/` + `version.txt`).

So the first start raises `FileNotFoundError`, and with `Restart = "always"` / `RestartSec = "1s"` it restarts every second indefinitely.

Suggestion: ship `conf/logging.json` with the Python package and seed the state dir from it in a `preStart` (or via `systemd.tmpfiles`). Making it a proper module option later would be nicer still.

### 2. `--tmp-folder /tmp` combined with `PrivateTmp = true` discards state that needs to persist

`PrivateTmp` gives the unit a fresh empty `/tmp` on every restart, and two things that must survive restarts live in the temp folder:

* **The session secret.** It defaults to `<temp_folder>/secret.dat` (`src/model/server_conf.py:220`). Losing it on every restart logs out every user and invalidates all previously issued file-download links.
* **Migration bookkeeping.** `migrations.txt` lives in the temp folder (`src/migrations/migrate.py:64,77`). Once the conf folder is non-empty, `_is_new_installation()` returns false and the marker file is gone, so *all* migrations re-run on every boot (`src/migrations/migrate.py:402-436`). Two of them write `context.conf_file` (`:205`, `:327`) — which here is a read-only `/etc` → Nix store symlink — so, for example, an LDAP setup with a legacy `username_pattern` would hit a write failure at startup.

Suggestion: point `--tmp-folder` at something under `StateDirectory` (e.g. `%S/script-server/temp`), or set `secret_storage_file` explicitly and keep `migrations.txt` persistent.

### 3. The module can't be used standalone

`package = pkgs.script-server` only resolves if the consumer *also* applies `overlays.default`. As it stands,

```nix
imports = [ inputs.script-server.nixosModules.default ];
services.script-server.enable = true;
```

fails to evaluate with `attribute 'script-server' missing`. Could you make `module.nix` a function of `self` and default to `self.packages.${pkgs.stdenv.hostPlatform.system}.script-server`, or have the module set `nixpkgs.overlays` itself?

## Security

### 4. `/etc/script-server/conf.json` is world-readable

`environment.etc` defaults to mode `0444`. The server config can hold the LDAP bind password, OAuth client secrets and SMTP credentials, so on a multi-user host every local user could read them. Worth `mode = "0400"` with an owner (a bit awkward under `DynamicUser`; `LoadCredential=` is the cleaner route).

Related: secrets inlined into `settings` end up in the world-readable Nix store. A `settingsFile` option (or a documented credentials escape hatch) plus a warning in the option `description` would help people avoid that footgun.

### 5. `DynamicUser = true` is a questionable fit here

script-server's whole purpose is executing configured scripts. With `DynamicUser` the UID rotates across boots, so anything those scripts write outside `StateDirectory`/`LogsDirectory` is orphaned. It also means the default script-config dir really lives at `/var/lib/private/script-server/conf` (mode 0700, root-owned) — which is exactly the directory an admin has to populate to get any value out of the server. A static user, or at minimum a note in the option description, would be friendlier.

### 6. No `port` / `openFirewall` / `user` options

The listening port is buried in free-form `settings`, so the module can't open the firewall or be introspected by other modules.

## Packaging

### 7. Optional runtime dependencies are missing

Only `tornado` is declared, but `ldap3` (`src/auth/auth_ldap.py`, imported lazily at `src/model/server_conf.py:234`) and `requests` (`src/communications/destination_http.py`) are needed for LDAP auth and HTTP alerts/callbacks respectively. Because both imports are lazy, users only find out at runtime, with an `ImportError`. Our `requirements.txt` has the same gap — but if you're up for adding them here (or exposing `withLdap` / `withHttp` flags), that'd be a nice improvement.

### 8. `bin/script-server` doesn't work outside the unit

`environment.systemPackages = [ cfg.package ]` puts a `script-server` binary on every user's `PATH` that immediately `sys.exit(-1)`s: `main()` derives `project_path` from `os.getcwd()`, and the `os.chdir(...)` only happens in the `if __name__ == '__main__'` guard (`src/main.py:159-163`), which the `main:main` console script never runs.

Wrapping the binary (`makeWrapper --run 'cd ${dist}'`) would fix that *and* let you drop the `WorkingDirectory` hack in the unit.

Heads-up while you're in there: the console-script path also skips `AsyncHTTPClient.configure(None, defaults=dict(validate_cert=False))` from that same guard. This PR is effectively the first consumer of the `main:main` entry point, so it's a code path we've never exercised.

### 9. `src/tests` gets installed

The fileset pulls in all of `./../src`, so setuptools src-layout auto-discovery will publish a top-level `tests` package into `site-packages`. Worth trimming the fileset (or excluding it in `pyproject.toml`). While you're there, `pythonImportsCheck = [ "main" ]` would at least prove the package imports.

### 10. Two sources of truth for the version

`package-dist.nix` hardcodes `version = "1.18.0"` while `package.nix` reads it from `pyproject.toml` — these will drift. `web-src/package.json` carries the same version and can be read with `builtins.fromJSON`.

### 11. `installPhase` drops the hooks

`runHook preInstall` / `runHook postInstall` are missing, which breaks `overrideAttrs` for downstream consumers.

### 12. `cp -r ../web "$out"` depends on an implementation detail

It relies on `vue.config.js` setting `outputDir: '../web'` (i.e. writing outside the source root). That works today, but it'll break silently if that config ever changes. A `[ -d ../web ]` guard, or passing the output dir explicitly, would make the failure loud.

### 13. Nothing keeps `npmDepsHash` / `flake.lock` honest

`npmDepsHash` is a manual digest of `web-src/package-lock.json`, so the first frontend dependency bump will silently break the build — and that breakage lands on whoever bumps it, not on the person who can fix the hash. Could you add a CI job that runs `nix build .#script-server-dist` (and ideally `nix flake check`)? For me that's close to a prerequisite for keeping this maintainable long-term.

### 14. `meta` is missing from `package.nix`

`description`, `homepage`, `license = lib.licenses.gpl3Only`, `platforms`, and especially `mainProgram = "script-server"` (needed for `nix run`).

## Minor

* `lib.mkEnableOption "Whether to enable script-server."` renders as *"Whether to enable Whether to enable script-server.."* — `mkEnableOption` already adds the prefix, so `mkEnableOption "script-server"` is enough.
* Option `package-dist` should be camelCase (`distPackage`) to match NixOS convention; kebab-case has to be quoted in consumer configs.
* `configuration` is a bit vague for "the script config directory" — maybe `configDir`, and `lib.types.path` rather than `str`.
* `systemd.services.script-server.enable = true;` is redundant inside `mkIf cfg.enable`.
* `.gitignore`: `/result` misses `result-*` (multi-derivation builds) — `/result*` covers both. `.direnv/` might be worth adding too.
* Nothing in the README mentions the flake, so nobody will discover it.

## One question about scope

The description advertises `nix run .#demo` and `nixosConfigurations.demo.config.system.build.images.*`, but I don't see a `demo` package or any `nixosConfigurations` in the diff — only `packages.{script-server,script-server-dist}` and `nixosModules.default`. Did those get left out intentionally (follow-up PR?), or is a file missing?

---

To be clear about the value here: I'm keen on this landing. Items 1–3 are what block it for me; the rest are mostly polish, and I'm happy to take several of them as follow-ups if you'd rather keep this PR focused.

bugy:
Hi @duesee since I have no idea about nixOS, I cheated and asked LLM for help with reviewing, feel free to address what makes sense and reject what doesn't
Sorry for delays, I'm very scarce on free evenings recently

duesee:
Don't worry, it's fine! Did you review that the information provided about script-server is roughly correct? There are some details there about script-server I didn't know. Besides that it all makes sense and I can keep working on it :-)

bugy:
@duesee to be honest all the comments look very nixos specific
I don't see anything related to a pure script server

duesee:
Thanks.

I'll go point by point, also limited time currently, sorry for the delay.

1. Fresh install will crash-loop — conf/logging.json is never provisioned

Yes. Actually, I have a specific "conf" package but forgot to define (or tell how to) here. Will change :+1:

2. --tmp-folder /tmp combined with PrivateTmp = true discards state that needs to persist

This is what I was referring to when I said "information provided about script-server". Is the session secret and migration stuff really in /tmp? If so, should it be in the state directory then, i.e., /var/lib/script-server/tmp? Does that sound correct to you?

3. The module can't be used standalone

~I am not sure I understand this. I'll try to make sense of it, maybe the setup can really be simplified.~ Got it. On it.

PR #824, opened 2026-07-31
Add pagination to history and admin logs

cpadlab:
I have added pagination to the "History" and "Logs" pages, as loading all entries at once was significantly slowing things down when there were many entries.

To improve performance, I also implemented an in-memory cache (`_entries_cache` in `logging.py`) so log files are parsed once on discovery and served directly from RAM without repeated disk reads.

I have tested this locally and added unit tests for both backend and frontend.

Key changes:
- Backend: Added pagination query params (`?page=X&size=Y`) and implemented RAM caching for logs. Optimized status checks to only run for active page entries.
- Frontend: Added a reusable Paginator component (10, 25, 50, 100, 250, 500 entries) and updated production build files.

### DIFF
--- a/src/execution/logging.py
+++ b/src/execution/logging.py
@@ -117,6 +117,7 @@ class ExecutionLoggingService:
 
         self._visited_files = set()
         self._ids_to_file_map = {}
+        self._entries_cache = {}
         self._output_loggers = {}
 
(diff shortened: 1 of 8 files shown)


cpadlab:
Superceded by #826 which implements backend cursor-based pagination and search/sort.

PR #826, opened 2026-08-05
Paginate execution history with cursors, indexed in the logging service

bugy:


bugy:
Hi @cpadlab inspired by your PR, I propse the similar functionality but implemented slightly differently
Could you check if this works well for your setup?

It also moves search/sort functionality to BE, so we don't have to load everything on the FE.

cpadlab:
Hi @bugy,
Thanks for taking the time to build on this! I like your approach much better. I tested your PR locally with our log setup and everything works smoothly. I've just closed my PR (#824) in favor of this one. Thanks again for the quick iteration and great work!

bugy:
Great, thanks for confirming @cpadlab

